Issue #4, opened 2017-04-13
Still link to b.p.o when both issue number and trivial label are found


Mariatta:
For example: https://github.com/python/cpython/pull/1120

It has both bpo-NNNN in the PR title and `trivial` label.
bedevere status check should still have details link to the said issue.

Thanks :)

terryjreedy:
Definitely.  In this case, the issue https://bugs.python.org/issue29861 and initial patch were not at all trivial, just the minor cleanup for occasional test failures reported on the issue.

PR #5, opened 2017-04-14
Link to b.p.o when both issue number and trivial label are found

Mariatta:
Fixes https://github.com/python/bedevere/issues/4

### DIFF
--- a/bedevere/bpo.py
+++ b/bedevere/bpo.py
@@ -35,7 +35,7 @@ async def set_status(gh, event):
         else:
             status = FAILURE_STATUS
     else:
-        status = get_found_issue_number_status(issue_number_found)
+        status = create_success_status(issue_number_found)
     await _post_status(gh, event, status)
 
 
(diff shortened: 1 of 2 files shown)


bedevere/bpo.py
```diff
@@ -55,11 +51,27 @@
...
 
 @router.route("pull_request", "unlabeled")
 async def removed_label(gh, event):
     """Re-check the status if the "trivial" label is removed."""
     if event.data["label"]["name"] == TRIVIAL_LABEL:
         await set_status(gh, event)
+
+
+def get_found_issue_number_status(found_issue):
+    """Get the status for when an issue number was found in the title"""
```
brettcannon: Missing a period and you're creating a status instead of getting one.

bedevere/bpo.py
```diff
@@ -55,11 +51,27 @@
...
 
 
 @router.route("pull_request", "unlabeled")
 async def removed_label(gh, event):
     """Re-check the status if the "trivial" label is removed."""
     if event.data["label"]["name"] == TRIVIAL_LABEL:
         await set_status(gh, event)
+
+
+def get_found_issue_number_status(found_issue):
```
brettcannon: You're actually not "getting" a status but creating one. So maybe `create_success_status()` (technically ambiguous due to trivial issues, but they have their own static status so I think it's fine)?

Mariatta: Thanks I renamed this :)

Mariatta:
🤔 Anything I need to do about codecov?

brettcannon:
@Mariatta don't worry about codecov; it was turned on after this PR was created so it's just in a weird state.

PR #9, opened 2017-04-14
Status check for reverting a commit

Mariatta:
If a PR starts with the word "Revert", then check if the
body contains the reason for the revert.

Closes https://github.com/python/core-workflow/issues/72

### DIFF
--- a/bedevere/revert_pr.py
+++ b/bedevere/revert_pr.py
@@ -1,0 +1,42 @@
+import re
+
+from . import router
+
+
+REVERT_RE = re.compile(r"^Revert")
+STATUS_TEMPLATE = {"context": "bedevere/revert-pr"}
+SUCCESS_STATUS = STATUS_TEMPLATE.copy()
+SUCCESS_STATUS["state"] = "success"
+SUCCESS_STATUS["description"] = "Found the reason to revert the PR."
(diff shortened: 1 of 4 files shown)


bedevere/revert_pr.py
```diff
@@ -0,0 +1,42 @@
+import re
+
+from . import router
+
+
+REVERT_RE = re.compile(r"^Revert")
+STATUS_TEMPLATE = {"context": "bedevere/revert-pr"}
```
brettcannon: This is too generic of a context. Maybe "bedevere/reason-to-revert"?

bedevere/revert_pr.py
```diff
@@ -0,0 +1,42 @@
...
+SUCCESS_STATUS = STATUS_TEMPLATE.copy()
+SUCCESS_STATUS["state"] = "success"
+SUCCESS_STATUS["description"] = "Found the reason to revert the PR."
```
brettcannon: "Reason for reverting the commit is specified."?

Mariatta:
Feel free to close if it's decided that we don't need a status check for this :) Thanks.

brettcannon:
Thanks for at least taking a stab at this, @Mariatta 😄

PR #24, opened 2017-06-26
Add a link to b.p.o as a comment

### LINKED ISSUE #3: Add the link to b.p.o as a comment on GitHub
Currently, the link back to b.p.o is only shown as a status check when PR is not yet merged.
Often I still re-read a merged PR, for example while backporting changes.
After a cherry-pick PR has been merged, we're supposed to go back to the bug tracker and close the issue.
However, since the PR has been merged, the status check is no longer visible.
So I think it will be nice if the link to the bug tracker can be added as a GitHub comment that is always going to be available.

What do you think about this?

kushaldas:
Adds the comment in the body of the PR.


bedevere/bpo.py
```diff
@@ -4,7 +4,16 @@
...
-
+re1='(<!--.*?-->)'	# HTML Comment 1
+re2='.*?'	# Non-greedy match on filler
+re3='((?:http|https)(?::\\/{2}[\\w]+)(?:[\\/|\\.]?)(?:[^\\s"]*))'	# HTTP URL 1
+re4='.*?'	# Non-greedy match on filler
+re5='(<!--.*?-->)'	# HTML Comment 2
+BPO_URL_RE = re.compile(re1+re2+re3+re4+re5,re.IGNORECASE|re.DOTALL)
+BPO_MSG = """{0}
+<!-- issue number -->
+{1}
```
brettcannon: Would it make sense to have a fancier message? E.g. "(Please only discuss the code in this pull request here on GitHub; all other discussion should occur on the issue at [bpo-{issue_number}](https://bugs.python.org/issue{issue_number}).)"

bedevere/bpo.py
```diff
@@ -37,6 +51,14 @@
         else:
             status = FAILURE_STATUS
     else:
+        if "body" in event.data["pull_request"]:
+            body = event.data["pull_request"]["body"]
+            bpo_url_found = BPO_URL_RE.search(body)
+            if not bpo_url_found:
+                issue_number = issue_number_found.group(1)
+                bpo_url = f"https://bugs.python.org/issue{issue_number}"
```
brettcannon: You could add this to `BPO_MSG` so that it's all part of the message template.

bedevere/bpo.py
```diff
@@ -37,6 +51,14 @@
         else:
             status = FAILURE_STATUS
     else:
+        if "body" in event.data["pull_request"]:
+            body = event.data["pull_request"]["body"]
+            bpo_url_found = BPO_URL_RE.search(body)
```
kushaldas: This is something similar I did at first (personally I don't use re module that much). I will use this msg text now.

bedevere/bpo.py
```diff
@@ -37,6 +51,14 @@
         else:
             status = FAILURE_STATUS
     else:
+        if "body" in event.data["pull_request"]:
```
kushaldas: Yes, there are edits (iirc) which does not have body, my test instance crashed due to this.

tests/test_bpo.py
```diff
@@ -205,3 +210,40 @@ def __init__(self, *, getitem=None):
...
+            "url": "https://api.github.com/repos/blah/blah/pulls/1347",
+            "title": "[3.6] bpo-1234: an issue!",
+            "body": """The body.\n<!-- issue-number: bpo-1234 -->\n"https://bugs.python.org/issue1234"\n<!-- /issue-number -->"""
+        },
+    }
+    event = sansio.Event(data, event="pull_request", delivery_id="12345")
+    gh = FakeGH()
+    await bpo.router.dispatch(event, gh)
+    assert gh.patchdata is None
+    assert gh.patchurl is None
```
brettcannon: Missing a newline at the end of the file.

bedevere/bpo.py
```diff
@@ -37,6 +44,17 @@
...
+            body = event.data["pull_request"]["body"]
+            if CLOSING_TAG not in body:
+                issue_number = issue_number_found.group("issue")
+                BPO_MSG = f"""{body}\n
+<!-- {TAG_NAME}: bpo-{issue_number} -->
```
brettcannon: To make this more readable, you can indent it and then use [`textwrap.dedent()`](https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent) to strip the leading whitespace.

bedevere/bpo.py
```diff
@@ -37,6 +44,17 @@
         else:
             status = FAILURE_STATUS
     else:
+        if "body" in event.data["pull_request"]:
+            body = event.data["pull_request"]["body"]
+            if CLOSING_TAG not in body:
+                issue_number = issue_number_found.group("issue")
+                BPO_MSG = f"""{body}\n
```
brettcannon: Might as well just have an actual newline in the triple-quoted string rather than a newline literal (I actually overlooked it when I was checking that there was a newline separating the body from the link).

tests/test_bpo.py
```diff
@@ -10,6 +10,8 @@ class FakeGH:
 
     def __init__(self, *, getitem=None):
         self._getitem_return = getitem
+        self.patchurl = None
```
brettcannon: Underscores have been used in the rest of the test code, so `patch_url` and `patch_data` for consistency.

brettcannon:
I just wanted to check in and make sure you're not blocked on me (no pressure, I just don't want to be a hold-up for you).

brettcannon:
I also had the time to address my own comments, so if you're happy with the resulting patch let me know and I will merge it!

brettcannon:
One thing I've noticed while playing with other GitHub stuff is they use `\r\n` for line separators in the body. Is that something we need to worry about when re-constructing the body?

kushaldas:
The examples I can see don't have ```\r\n```
Like

```
"body": "Can we merge this without testing?\n<!-- issue-number: bpo-30442 -->\nhttps://bugs.python.org/issue30442\n<!-- /issue-number -->\n",
```

kushaldas:
The commits look good, you can go ahead and merge. I got stuck in something else for the last 2 weeks :(

bedevere/bpo.py
```diff
@@ -37,6 +40,19 @@
         else:
             status = FAILURE_STATUS
     else:
+        if "body" in event.data["pull_request"]:
+            body = event.data["pull_request"]["body"]
+            if CLOSING_TAG not in body:
+                issue_number = issue_number_found.group("issue")
+                BPO_MSG = textwrap.dedent(f"""\
```
berkerpeksag: Unfortunately, `textwrap.dedent()` doesn't work with f-strings (or there is some other reason I don't know) See https://github.com/python/bedevere/pull/27/files/dcc17d64f1c9a9d6b65703abbf697c132d76d785#r127503862 for the previous discussion.

You can see the example in my latest PR: https://github.com/python/cpython/pull/2807

brettcannon: Already have a fix waiting for travis to clear.

berkerpeksag: Thanks!

brettcannon: And it's merged, so once Travis is green for master again it will go live.

And I don't know why textwrap.dedent() isn't working since I don't see how an f-string would influence the outcome in anyway. My guess is the string as being inserted has some unexpected whitespace that's throwing the algorithm off.

Mariatta: Is this a bug with textwrap.dedent() or f-strings?
Worth opening an issue in b.p.o about this?

brettcannon: Nope, I see what happened. Berker's body had newlines in it that led to parts of the body not being equally indented like the string literal, so when `textwrap.dedent()` got it there was whole chunks of text not indented so it assumed there was nothing to dedent. IOW it worked as intended and it just  didn't occur to me when doing the code review that the inserted body would have newlines followed by text that wouldn't be indented.

PR #33, opened 2017-07-22
Set a stage label for a PR

brettcannon:


bedevere/awaiting.py
```diff
@@ -0,0 +1,219 @@
...
+    # single pull request, but just in case there are we might as well clean
+    # up the situation when we come across it.
+    for label in issue["labels"]:
+        stale_name = label["name"]
+        if stale_name.startswith(LABEL_PREFIX + " "):
+            await gh.delete(issue["labels_url"], {"name": stale_name})
+    await gh.post(issue["labels_url"], data=[label_name])
+
+
+async def is_core_dev(gh, username):
```
berkerpeksag: I wrote a similar function for my cherry-pick bot (only accept `/cherry-pick` commands from core devs), but GitHub API's rate limit was an issue so I end up saving team data into a JSON file. Did you have the same problem?

brettcannon: I hope not. 😁 Gidgethub has a caching mechanism so that as long as the requests return the proper headers then it shouldn't lead to that many more calls that count against the API limit.

brettcannon: There's also the option of using https://developer.github.com/v3/orgs/teams/#list-team-members which should also minimize the amount of calls necessary (although I still bet the caching will do the trick since Bedevere is set to cache over 200 MB of cached data).

brettcannon:
To help with coverage I need to stop being clever with pull request versus issue and just have helper functions that do what's necessary. E.g. have `stage()` just take an issue and if I'm working from a PR then use a `pr_to_issue()` function as an argument to the `stage()` call. Same goes for just passing in the URL for `core_dev_reviewers()`.

PR #34, opened 2017-07-25
set the "maintainer_can_modify" flag to True in bpo.py

Mariatta:
When bedevere updates the PR body, also pass in the `"maintainer_can_modify"` flag.

I'm wondering if it gets set to `False` when the flag is not set, which is causing recent PRs in CPython repo to not be editable by maintainers. Just a guess. Maybe a better approach is to first check what the flag was set to before updating the PR.


brettcannon:
Sorry for missing this PR, @Mariatta ! I did the quick fix in the browser and I'm pressed for time at work today so I didn't check my GitHub notifications before submitting my own PR.

Mariatta:
No prob :)

PR #66, opened 2017-10-08
Support a more muted trigger phrase for another review

brettcannon:
See https://mail.python.org/pipermail/python-committers/2017-October/004902.html and https://mail.python.org/pipermail/core-workflow/2017-October/001039.html for reference.

The old trigger phrase is left in as both an easter egg and for supporting people who have begun to use the phrase out of habit. The easter egg in the comment about the trigger phrase is untouched.

### DIFF
--- a/tests/test_stage.py
+++ b/tests/test_stage.py
@@ -284,7 +284,7 @@ async def test_new_review():
     assert labeling[1] == [awaiting.Blocker.changes.value]
     message = gh.post_[1]
     assert message[0] == "https://api.github.com/comment/42"
-    assert awaiting.REQUEST_CHANGE_REVIEW in message[1]["body"]
+    assert awaiting.BORING_TRIGGER_PHRASE in message[1]["body"]
 
     # Comment reviews do nothing.
     data = {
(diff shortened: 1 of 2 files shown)


tests/test_stage.py
```diff
@@ -408,6 +408,35 @@ def __init__(self, *, getiter=None, getitem=None, delete=None, post=None):
     assert "@gvanrossum" in comment_body
     assert "not-core-dev" not in comment_body
 
+    # All is right with the Monty Python world.
```
merwok: Isn’t this a different unit test?

brettcannon: If you want to get technical, sure. 😉 But there's a lot of reused setup code so it's not an isolated test in terms of actual code. Plus the whole bot is stateless by design so this won't influence outcomes by having this here.

merwok: Thanks for explaining!  Good to know if I contribute to or fork your bot :)

brettcannon: I have no problem if someone wanted to do a better separation of test code, I just wanted to get this done sooner rather than later so I didn't take the time to extract common code for a refactoring. 😄

brettcannon:
Just an FYI, I expanded the trigger message to be "I have made the requested changes; please review again". I figured it wouldn't hurt to make sure people at least paste in a confirmation that they looked at the review. 😉

PR #67, opened 2017-10-09
 Periodic job to label stale PRs python/core-workflow#93

chetankm-cs:
This pr adds a periodic job to detect and label stale pull requests.

A pull request is considered stale if it was not updated in last 30 days and it has labels 
CLA not signed/ awaiting changes.

The job will run daily at 12am and will mark pull request stale only if required.
The job is scheduled using https://github.com/agronholm/apscheduler.
Apscheduler is a pure python library and doesn't depend on any external cron/task queue for triggering the task. This helps in making the bot cross platform.

### DIFF
--- a/bedevere/__main__.py
+++ b/bedevere/__main__.py
@@ -10,14 +10,13 @@ import cachetools
 from gidgethub import aiohttp as gh_aiohttp
 from gidgethub import routing
 from gidgethub import sansio
+from apscheduler.schedulers.asyncio import AsyncIOScheduler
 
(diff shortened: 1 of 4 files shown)


Mariatta:
Thanks for the work on this. Sorry we haven't been able to follow up for a few years :( 
Adding and closing "stale" PRs can now be done through GitHub Actions. We have an open PR in CPython to address it (https://github.com/python/cpython/pull/21247) but we haven't reached agreement on whether we want this functionality or not 🤷🏻‍♀️

Issue #15, opened 2017-05-26
Validate the bpo number


Mariatta:
The bpo number provided in https://github.com/python/cpython/pull/1798 is not valid (46895)
 https://bugs.python.org/issue46895 is a 404

Maybe bedevere can also validate whether the issue number is correct :)

brettcannon:
Yeah, we could do a simple GET request and see if the response is a 404 or not to flag the issue number as formatted properly but not an actual issue number.

serhiy-storchaka:
Doing a HEAD request may be enough.

lysnikolaou:
Is this issue resolved or should it be implemented? If so, I would be happy to take it up.

lysnikolaou:
This should be closed since #159 is now merged.

PR #72, opened 2017-11-09
validate bpo number

levlaz:
Add helper function to check if the issue number exists on bpo.

set FAILURE_STATUS in the event that the issue does not exist.

Fix https://github.com/python/bedevere/issues/15

### DIFF
--- a/bedevere/bpo.py
+++ b/bedevere/bpo.py
@@ -1,5 +1,7 @@
 """Check if a bugs.python.org issue number is specified in the pull request's title."""
 import re
+import urllib.request
+import urllib.error
 
 from gidgethub import routing
 
@@ -21,7 +23,7 @@ ISSUE_RE = re.compile(r"bpo-(?P<issue>\d+)")
 SKIP_ISSUE_LABEL = util.skip_label("issue")
 STATUS_CONTEXT = "bedevere/issue-number"
 # Try to keep descriptions at or below 50 characters, else GitHub's CSS will truncate it.
-_FAILURE_DESCRIPTION = 'No issue # in title or "skip issue" label found'
+_FAILURE_DESCRIPTION = 'Issue does not exit, No issue # in title or "skip issue" label found'
 _FAILURE_URL = "https://devguide.python.org/pullrequest/#submitting"
 FAILURE_STATUS = util.create_status(STATUS_CONTEXT, util.StatusState.FAILURE,
                                     description=_FAILURE_DESCRIPTION,
@@ -48,7 +50,10 @@ async def set_status(event, gh, *args, **kwargs):
                 new_body = BODY.format(body=body, issue_number=issue_number)
                 body_data = {"body": new_body, "maintainer_can_modify": True}
                 await gh.patch(event.data["pull_request"]["url"], data=body_data)
-        status = create_success_status(issue_number_found)
+        if validate_issue_number(issue_number_found.group("issue")):
+            status = create_success_status(issue_number_found)
+        else:
+            status = FAILURE_STATUS
     await util.post_status(gh, event, status)
 
 
@@ -89,3 +94,13 @@ def create_success_status(found_issue):
     return util.create_status(STATUS_CONTEXT, util.StatusState.SUCCESS,
                               description=f"Issue number {issue_number} found",
                               target_url=url)
+
+def validate_issue_number(issue_number):
+    """Validate that the issue number exists on bpo"""
+    url = f"https://bugs.python.org/issue{issue_number}"
+    req = urllib.request.Request(url=url, method='HEAD')
+    try:
+        urllib.request.urlopen(req)
+        return True
+    except urllib.error.HTTPError:
+        return False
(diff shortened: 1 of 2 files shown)


bedevere/bpo.py
```diff
@@ -21,7 +23,7 @@
 SKIP_ISSUE_LABEL = util.skip_label("issue")
 STATUS_CONTEXT = "bedevere/issue-number"
 # Try to keep descriptions at or below 50 characters, else GitHub's CSS will truncate it.
-_FAILURE_DESCRIPTION = 'No issue # in title or "skip issue" label found'
+_FAILURE_DESCRIPTION = 'Issue does not exit, No issue # in title or "skip issue" label found'
```
levlaz: This felt a little dirty, but I was not sure if we wanted to "yet another" edge case of failure description. Happy to refactor this if we want to be more specific of the reason for failure.

tests/test_bpo.py
```diff
@@ -267,3 +290,11 @@ def __init__(self, *, getitem=None):
     await bpo.router.dispatch(event, gh)
     assert gh.patch_data is None
     assert gh.patch_url is None
+
+
+def test_validate_issue_number():
```
levlaz: This is a "unit" test but it actually makes a call to bpo... maybe this should be mocked as well (or it may not be necessary all together).

bedevere/bpo.py
Mariatta: Hmm, this might be too long as a status check message.
I'm thinking it can be a separate status check, maybe "bedevere/valid-issue"? @brettcannon, what do you think?

bedevere/bpo.py
```diff
@@ -89,3 +94,13 @@ def create_success_status(found_issue):
     return util.create_status(STATUS_CONTEXT, util.StatusState.SUCCESS,
                               description=f"Issue number {issue_number} found",
                               target_url=url)
+
+def validate_issue_number(issue_number):
+    """Validate that the issue number exists on bpo"""
```
Mariatta: The docstring should end with a period :)

bedevere/bpo.py
```diff
@@ -89,3 +94,13 @@ def create_success_status(found_issue):
...
+    url = f"https://bugs.python.org/issue{issue_number}"
+    req = urllib.request.Request(url=url, method='HEAD')
+    try:
+        urllib.request.urlopen(req)
+        return True
+    except urllib.error.HTTPError:
+        return False
```
Mariatta: Needs a newline.

tests/test_bpo.py
Mariatta: My preference is to mock the call to b.p.o.

bedevere/bpo.py
```diff
@@ -1,5 +1,7 @@
 """Check if a bugs.python.org issue number is specified in the pull request's title."""
 import re
+import urllib.request
```
Mariatta: How about `from urllib import error, request` instead?

levlaz:
Thank you for the review @Mariatta, I made all of the requested changes except for defining a new status check message since you want to get some feedback from Brett. 

Happy to implement whatever you all decide.

bedevere/bpo.py
brettcannon: I don't think we need another status check (we are already at 4). Is there any reason the message can't be dynamically specified like for the `NEWS.d` check? In this instance the current message is the one that shows up if anything is found, but if an issue number is discovered but found not to exist, a different message saying something like "issue #NNNN not found at bugs.python.org" or something would be set while still flagging a status check failure.

Make sense?

Mariatta: Using a different message sounds good :)

bedevere/bpo.py
```diff
@@ -89,3 +92,13 @@ def create_success_status(found_issue):
     return util.create_status(STATUS_CONTEXT, util.StatusState.SUCCESS,
                               description=f"Issue number {issue_number} found",
                               target_url=url)
+
+def validate_issue_number(issue_number):
```
brettcannon: Start with an underscore since this shouldn't be used outside of this module.

bedevere/bpo.py
terryjreedy: Corrections to message: /exit/exist/, /No/no/, /title/title,/.  But +1 to situation specific message.   This is not an exception raised in inner loops for control flow.  (The reason for some quick and vague exception messages.)

bedevere/bpo.py
```diff
@@ -89,3 +92,13 @@ def create_success_status(found_issue):
...
+    """Validate that the issue number exists on bpo."""
+    url = f"https://bugs.python.org/issue{issue_number}"
+    req = request.Request(url=url, method='HEAD')
+    try:
+        res = request.urlopen(req)
```
merwok: Shouldn’t this use an async request instead of dear old urllib?

Mariatta:
@levlaz still working on this?

lysnikolaou:
Since @levlaz is apparently no longer working on this, should this PR maybe a closed? @Mariatta?

brettcannon:
Due to this stalling out, I'm closing the PR.

PR #75, opened 2017-11-23
Make the "Change requested" message brief when PR is authored by Core Dev

Mariatta:
Follow up from [core-workflow](https://mail.python.org/mm3/archives/list/core-workflow@python.org/thread/H5CRUS6AHTTJDXWUAMJZWKXEAMMRQPM3/) email thread.

The change requested message is now shorter when PR author is a core dev:
```
When you're done making the requested changes, leave the comment:
`I have made the requested changes; please review again`
```

### DIFF
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -70,6 +70,14 @@ you're ready for them to take another look at this pull request.
 {{easter_egg}}
 """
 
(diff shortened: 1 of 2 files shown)


Mariatta:
Thanks for the review! 😄

Issue #71, opened 2017-11-07
Only remove "awaiting change" PR label after all reviewers approved.


Mariatta:
When there are multiple reviewers who requested changes, the "awaiting change" label should only be removed after all reviewers approved the new changes.

As brought up by @1st1 in https://github.com/python/cpython/pull/4314:

> core-dev A requests changes; bedevere-bot assigns an awaiting changes label.
> core-dev B approves the PR; bedevere-bot removes the awaiting changes label.
> The label should only be removed when all reviewers clear the PR.

brettcannon:
So I debated with myself about this when I initially implemented the feature. The problem is that core devs are not the most responsive folks in the world due to lack of time. So if two or more people review and just one of them delays, it would then delay the whole PR. This also deviates somewhat from the old workflow where you didn't have to wait for other core devs to sign off if the reviewer clearing the PR thought the other comments were appropriately taken care of.

IOW I don't know if this is a good or bad thing to do, hence why I went with the one that was the easiest to code up. 😁

Mariatta:
I don't personally have strong preference. The simpler workflow is fine too :) 
Maybe @1st1 want to reconsider this scenario, if it's fine to remove the awaiting change label as long as at least one core dev approves it?

brettcannon:
I think it's worked well enough this entire time that we should drop this idea.

PR #76, opened 2017-11-24
Only remove "awaiting change" PR label after all reviewers approved.

Mariatta:
When there are multiple reviewers who requested changes, the "awaiting change" label should only be removed after all reviewers approved the new changes.

Closes https://github.com/python/bedevere/issues/71


brettcannon:
I'm going to close this for now, but if we decide we do want this change to the workflow we can obviously open it again.

PR #78, opened 2017-12-07
Use Travis to test against 3.6-dev and 3.7-dev

Mariatta:


brettcannon:
If we test against 3.6-dev then we can drop the 3.6 testing.

PR #79, opened 2018-01-24
Fix issue where PR without any description has its title edited

Mariatta:
This came up in https://github.com/python/cpython/pull/5291

The PR did not have any description, and the title was
edited to include the bpo number.
It caused a 500 error, because bedevere did not anticipate
this situation.

Updated the code so that it takes into consideration when the
body is empty.

### DIFF
--- a/bedevere/bpo.py
+++ b/bedevere/bpo.py
@@ -43,7 +43,7 @@ async def set_status(event, gh, *args, **kwargs):
     else:
         if "body" in event.data["pull_request"]:
             body = event.data["pull_request"]["body"]
-            if CLOSING_TAG not in body:
+            if not body or CLOSING_TAG not in body:
                 issue_number = issue_number_found.group("issue")
                 new_body = BODY.format(body=body, issue_number=issue_number)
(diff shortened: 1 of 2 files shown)


Mariatta:
Thanks @brettcannon for the quick review! 
But tests failed in 3.7 😥 ironically about asyncio stuff which I have no idea about @1st1 
https://travis-ci.org/python/bedevere/jobs/332943652#L568

brettcannon:
Looks like the version of `pytest-asyncio` is too old to be compatible with the new `async` keyword change in 3.7.

@Mariatta Can you update the `pytest-asyncio` dependency in your PR and see if that fixes it? (Also feel free to update any other dependencies you want.)

brettcannon:
I also just turned on pyup.io on this repo to help prevent this issue going forward.

brettcannon:
Looks like pyup.io pushed #81 which should hopefully fix this.

Mariatta:
Thanks! I'll rebase to get the latest changes.

brettcannon:
#81 seems to do the trick, so probably just a merge off of `master` now will do the trick.

Issue #100, opened 2018-03-01
adopt to aiohttp 3


asvetlov:
The library has changed testing API.
The bot uses non-public attributes, that's why tests fail.

I could find a time for working on issue on next week if nobody will pickup the task first

brettcannon:
I assume you're referencing #99 ?

asvetlov:
Yes, exactly

brettcannon:
Look forward to your PR since I will probably be writing a GH bot at work next week and I want to do it right 😄

asvetlov:
Fixed by #101

Mariatta:
Thanks!

asvetlov:
Sure!

PR #101, opened 2018-03-12
Adopt to aiohttp 3.0

asvetlov:
Fix for #100

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,5 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+.pytest_cache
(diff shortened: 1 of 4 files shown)


Mariatta:
Tests still failing on 3.7 ...

asvetlov:
Because I've added a strict check for raised deprecation warnings, I believe the change is for good intention.
Otherwise we'll stuck with unsupported anymore API change.

Please give me a time for fixing upstreams to not raise these annoying warnings.

asvetlov:
All done

Issue #106, opened 2018-05-31
Automatically add type labels


csabella:
The labels for `type-documentation` and `type-tests` could automatically be added to pull requests based on the types of files included.  For example, `type-documentation` would contain all `.rst` files, while `type-tests` would contain all files that start with `test_` (except for the blurb).  It may not catch all the changes for documentation and tests, but it would identify some, making triage of pull requests a bit easier.  The other types (bugfix or enhancement) might be able to be retrieved from the bpo.

I started working on this, but I ran into a question.  Currently `check_news` reads through the files to look for the blurb.  Since I plan on looking at all the file names as well, I was wondering if the code to get all the files from the PR should be moved to a level above `check_news` and the new function.  

For example in new `files.py`:
``` python
@router.register('pull_request', action='opened')
async def file_checks(event, gh, *args, **kwargs):
    pr_number = event.data['number']
    filelist = get_files_from_PR(number)
    await check_news(filelist)
    await check_type(filelist)
```

instead of retrieving the files twice:
```  python
@router.register('pull_request', action='opened')
async def check_news(event, gh, *args, **kwargs):
    pr_number = event.data['number']
    files_url = f'/repos/python/cpython/pulls/{pr_number}/files'
    async for file_data in gh.getiter(files_url):
        filename = file_data['filename']
        etc

@router.register('pull_request', action='opened')
async def check_type(event, gh, *args, **kwargs):
    pr_number = event.data['number']
    files_url = f'/repos/python/cpython/pulls/{pr_number}/files'
    async for file_data in gh.getiter(files_url):
        filename = file_data['filename']
        etc
```
I don't know what the pitfalls would be of trying to get all the files upfront, especially with this comment in check_news.
```
    # Could have collected all the filenames in an async set comprehension,
    # but doing it this way potentially minimizes the number of API calls.
```

brettcannon:
Factoring the code out is no problem. As for the comment, that's just me being perpetually paranoid about our quota. 😄

Mariatta:
Closed in https://github.com/python/bedevere/pull/108

PR #108, opened 2018-06-08
Add type labels for docs and tests.

csabella:
For issue #106 

- Create `files.py` as the route register for pull requests to get all the files and send them to `news.py` (blurb) and `prtype.py` (label type)
- Update `news.checknews()` to not have a route registry, but to use the file list from the parameters.
- Create `prtype.py` to assign type labels based on the files attached to a PR.
- Add tests for all of the above.

### DIFF
--- a/bedevere/files.py
+++ b/bedevere/files.py
@@ -1,0 +1,20 @@
+"""Checks related to files on a pull request."""
+import gidgethub.routing
+
+from .news import check_news
+from .prtype import classify_by_files
+from . import util
+
+
+router = gidgethub.routing.Router()
+
+
+@router.register('pull_request', action='opened')
+@router.register('pull_request', action='synchronize')
+async def retrieve_files(event, gh, *args, **kwargs):
+    pull_request = event.data['pull_request']
+    filenames = await util.filenames_for_PR(gh, pull_request)
+    await check_news(gh, pull_request, filenames)
+    if event.data['action'] == 'opened':
+        await classify_by_files(gh, pull_request, filenames)
+
--- a/bedevere/news.py
+++ b/bedevere/news.py
@@ -15,7 +15,6 @@ create_status = functools.partial(util.create_status, 'bedevere/news')
 
 DEVGUIDE_URL = 'https://devguide.python.org/committing/#what-s-new-and-news-entries'
 
-NEWS_NEXT_DIR = "Misc/NEWS.d/next/"
 FILENAME_RE = re.compile(r"""# YYYY-mm-dd or YYYY-mm-dd-HH-MM-SS
                              \d{4}-\d{2}-\d{2}(?:-\d{2}-\d{2}-\d{2})?\.
                              bpo-\d+(?:,\d+)*\.       # Issue number(s)
@@ -28,20 +27,12 @@ SKIP_LABEL_STATUS = create_status(util.StatusState.SUCCESS,
                                   description='"skip news" label found')
 
 
-@router.register('pull_request', action='opened')
-@router.register('pull_request', action='synchronize')
-async def check_news(event, gh, *args, **kwargs):
-    pr_number = event.data['number']
-    pull_request = event.data['pull_request']
-    # For some unknown reason there isn't any files URL in a pull request
-    # payload.
-    files_url = f'/repos/python/cpython/pulls/{pr_number}/files'
-    # Could have collected all the filenames in an async set comprehension,
-    # but doing it this way potentially minimizes the number of API calls.
+async def check_news(gh, pull_request, filenames=None):
+    if not filenames:
+        filenames = await util.filenames_for_PR(gh, pull_request)
     in_next_dir = file_found = False
-    async for file_data in gh.getiter(files_url):
-        filename = file_data['filename']
-        if not filename.startswith(NEWS_NEXT_DIR):
+    for filename in filenames:
+        if not util.is_news_dir(filename):
             continue
         in_next_dir = True
         file_path = pathlib.PurePath(filename)
@@ -58,7 +49,7 @@ async def check_news(event, gh, *args, **kwargs):
             status = SKIP_LABEL_STATUS
         else:
             if not in_next_dir:
-                description = f'No news entry in {NEWS_NEXT_DIR} or "skip news" label found'
+                description = f'No news entry in {util.NEWS_NEXT_DIR} or "skip news" label found'
             elif not file_found:
                 description = "News entry not in an appropriate directory"
             else:
@@ -81,4 +72,5 @@ async def label_removed(event, gh, *args, **kwargs):
     if util.no_labels(event.data):
         return
     elif util.label_name(event.data) == SKIP_NEWS_LABEL:
-        await check_news(event, gh)
+        pull_request = event.data['pull_request']
+        await check_news(gh, pull_request)
--- a/bedevere/news.py
+++ b/bedevere/news.py
@@ -28,6 +28,10 @@ SKIP_LABEL_STATUS = create_status(util.StatusState.SUCCESS,
 
 
 async def check_news(gh, pull_request, filenames=None):
+    """Check for a news entry.
+
+    The routing is handled through the filepaths module.
+    """
     if not filenames:
         filenames = await util.filenames_for_PR(gh, pull_request)
     in_next_dir = file_found = False
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -49,5 +49,4 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         await add_category(gh, issue, Category.tests)
     elif docs:
         await add_category(gh, issue, Category.documentation)
-    else:
-        return
+    return
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -25,8 +25,14 @@ async def add_category(gh, issue, category):
     await gh.post(issue["labels_url"], data=[category.value])
 
 
-async def classify_by_files(gh, pull_request, filenames):
-    """Categorize pull request."""
+async def classify_by_filepaths(gh, pull_request, filenames):
+    """Categorize the pull request based on the files it has modified.
+
+    If any paths are found which do not fall within a specific classification,
+    then no new label is applied.
+
+    The routing is handled by the filepaths module.
+    """
     issue = await util.issue_for_PR(gh, pull_request)
     docs = tests = 0
     for filename in filenames:
@@ -41,6 +47,5 @@ async def classify_by_files(gh, pull_request, filenames):
             return
     if tests:
         await add_category(gh, issue, Category.tests)
-        return
-    if docs:
+    elif docs:
         await add_category(gh, issue, Category.documentation)
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -1,0 +1,46 @@
+"""Label a pull request based on its type."""
+import enum
+import pathlib
+
+from . import util
+
+LABEL_PREFIX = "type"
+
+
+@enum.unique
+class Category(enum.Enum):
+    """Category of Pull Request."""
+    bugfix = f"{LABEL_PREFIX}-bugfix"
+    documentation = f"{LABEL_PREFIX}-documentation"
+    enhancement = f"{LABEL_PREFIX}-enhancement"
+    performance = f"{LABEL_PREFIX}-performance"
+    security = f"{LABEL_PREFIX}-security"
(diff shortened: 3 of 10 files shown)


bedevere/files.py
```diff
@@ -0,0 +1,20 @@
+"""Checks related to files on a pull request."""
```
brettcannon: "file paths in a pull request". Might want to also rename the file since nothing related to this file is actually looking inside the files.

bedevere/files.py
```diff
@@ -0,0 +1,20 @@
...
+router = gidgethub.routing.Router()
+
+
+@router.register('pull_request', action='opened')
+@router.register('pull_request', action='synchronize')
+async def retrieve_files(event, gh, *args, **kwargs):
+    pull_request = event.data['pull_request']
+    filenames = await util.filenames_for_PR(gh, pull_request)
+    await check_news(gh, pull_request, filenames)
+    if event.data['action'] == 'opened':
```
brettcannon: Why restrict this to only the initial set of files? What happens if someone gets asked to make a change that expands the scope of the PR?

bedevere/news.py
```diff
@@ -28,20 +27,12 @@
...
-@router.register('pull_request', action='synchronize')
-async def check_news(event, gh, *args, **kwargs):
-    pr_number = event.data['number']
-    pull_request = event.data['pull_request']
-    # For some unknown reason there isn't any files URL in a pull request
-    # payload.
-    files_url = f'/repos/python/cpython/pulls/{pr_number}/files'
-    # Could have collected all the filenames in an async set comprehension,
-    # but doing it this way potentially minimizes the number of API calls.
+async def check_news(gh, pull_request, filenames=None):
```
brettcannon: Add a comment that the route is handled through the `files` module.

bedevere/files.py
```diff
@@ -0,0 +1,20 @@
...
+from .prtype import classify_by_files
+from . import util
+
+
+router = gidgethub.routing.Router()
+
+
+@router.register('pull_request', action='opened')
+@router.register('pull_request', action='synchronize')
+async def retrieve_files(event, gh, *args, **kwargs):
```
brettcannon: This is doing more than just retrieving the files list, so the function name is misleading. Maybe `check_file_paths()`?

bedevere/prtype.py
```diff
@@ -0,0 +1,46 @@
...
+        filepath = pathlib.PurePath(filename)
+        if filepath.suffix == '.rst':
+            docs += 1
+        elif filepath.name.startswith('test_'):
+            tests += 1
+        else:
+            return
+    if tests:
+        await add_category(gh, issue, Category.tests)
+        return
```
brettcannon: Make this an `if`/`elif` and drop the `return`.

bedevere/prtype.py
```diff
@@ -0,0 +1,46 @@
...
+
+async def add_category(gh, issue, category):
+    """Apply this type label if there aren't any type labels on the PR."""
+    if any(label.startswith("type") for label in util.labels(issue)):
+        return
+    await gh.post(issue["labels_url"], data=[category.value])
+
+
+async def classify_by_files(gh, pull_request, filenames):
+    """Categorize pull request."""
```
brettcannon:
```
"""Categorize the pull request based on the files it has modified.

If any paths are found which do not fall within a specific classification, then no new label is applied.

"""
```

bedevere/prtype.py
```diff
@@ -0,0 +1,46 @@
...
+    issue = await util.issue_for_PR(gh, pull_request)
+    docs = tests = 0
```
brettcannon: Why the accumulation? Were you thinking of applying the label based on which type of file it had the most of?

bedevere/prtype.py
```diff
@@ -0,0 +1,46 @@
+"""Label a pull request based on its type."""
```
brettcannon: I'm struggling with whether this needs to be in its own file or should be moved to `files`. Up until this point every thematic thing that Bedevere did was contained in a single file, but `files` actually spreads across 3 at this point. I'm thinking it might be better to move this into `files` and then either do the same for `news.check_news()` or at least leave a comment as to why that one check breaks the overall rule.

tests/test_files.py
```diff
@@ -0,0 +1,182 @@
...
+    assert len(gh.post_url) == 2
+    assert gh.post_url[0] == 'https://api.github.com/some/status'
+    assert gh.post_data[0]['state'] == 'success'
+    assert gh.post_url[1] == 'https://api.github.com/some/label'
+    assert gh.post_data[1] == ['type-tests']
+
+
+async def test_news_and_tests():
+    filenames = [{'filename': '/path/to/docs.rst'}, {'filename': 'test_docs2.py'},
+                 {'filename': 'Misc/NEWS.d/next/Lib/' + GOOD_BASENAME}]
```
brettcannon: f-string 😄 (and anywhere else it makes sense.)

tests/test_news.py
```diff
@@ -57,13 +57,14 @@ def test_date_only(self):
         'action': 'opened',
         'number': 1234,
         'pull_request': {
+            'url': 'https://api.github.com/repos/cpython/python/pulls/1234',
```
brettcannon: Probably doesn't really matter, but it's `python/cpython`.

tests/test_news.py
```diff
@@ -57,13 +57,14 @@ def test_date_only(self):
...
             'statuses_url': 'https://api.github.com/some/status',
             'issue_url': 'https://api.github.com/repos/cpython/python/issue/1234',
         },
     }
     event = sansio.Event(event_data, event='pull_request', delivery_id=1)
-    await news.router.dispatch(event, gh)
-    assert gh.getiter_url == '/repos/python/cpython/pulls/1234/files'
+    await news.check_news(gh, event_data['pull_request'])
+    assert gh.getiter_url == 'https://api.github.com/repos/cpython/python/pulls/1234/files'
```
brettcannon: Any reason for adding the domain?

csabella: When I ran `pytest`, the test was failing the other way.  I think it was because of how I changed the iteration.  I wanted to pull the repo name from the JSON instead of hard-coding it for when I was trying to run this over my own repo during testing.  `news.py` originally had this line:
``` python
files_url = f'/repos/python/cpython/pulls/{pr_number}/files'
```
which I made as the following in `util.py`
``` python
    files_url = f'{pull_request["url"]}/files'
```

tests/test_news.py
csabella: I know -- but the existing test files (like test_news) had it like this.  Not sure why...   I thought maybe it was a 'just in case' so nothing happened at github.

tests/test_files.py
csabella: Of course!  I had copied this from the other tests and didn't even think about it.  :-)

bedevere/prtype.py
csabella: Thanks - I had it that way at first and wasn't sure which was the better way to write it.

bedevere/prtype.py
csabella: I initially was checking for blurbs a different way and was subtracting 1 if an `.rst` file was a blurb.  I left it as a count because I was thinking that the number of labels might grow.  I know it's not good to plan for 'what if' scenarios, but I was thinking about maybe adding a `size` label (xs, s, m, l, xl) would help with reviews.  If someone was looking through the PRs, and there was an xs doc issue, they might be able to review it quickly without having to click on every one to see which ones might be able to be merged quickly.  Especially since labels can be used as filters.

bedevere/prtype.py
csabella: Thanks!

bedevere/prtype.py
csabella: Yes, I wasn't sure how far I should go with restructuring this.  One reason I thought of having this in its own module is if we were able to add the other type labels by looking at the bug tracker.  So, for example, another routing entry and function could be added here to retrieve the bpo info and classify to `type-enhancement` or `type-bugfix` if the bug tracker had that as the type.  That lookup wouldn't make sense in the `file(path)` module and would start getting more like the Ni bot looking for the CLA.

bedevere/news.py
csabella: Added here and in `prtype`.

bedevere/files.py
csabella: I restricted this because of the ability to add and remove labels manually.  I thought that if someone really wanted to remove a label, then the bot might keep trying to add it back.  I didn't know if it would be better to fix it later if someone complained or just start out by only adding labels to the initial pull.

bedevere/files.py
csabella: Thanks!

bedevere/files.py
```diff
@@ -0,0 +1,20 @@
+"""Checks related to files on a pull request."""
+import gidgethub.routing
+
+from .news import check_news
```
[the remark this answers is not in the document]
csabella: Done.

bedevere/files.py
csabella: Good point.  I've renamed `files` to filepaths`.

tests/test_news.py
brettcannon: Or I just screwed up typing it the first time. Don't assume someone else didn't mess up before you. 😉

bedevere/prtype.py
brettcannon: Nice idea, but yeah, don't plan for the future. 😄 I keep reading the code and thinking, "where is the numeric value used?" So just drop it for now and if the desire to bring back is there it's easy enough to do. Always try to optimize for readability as that makes it more maintainable (and the future might lead to a different insight on how to measure size which could negate this initial work).

brettcannon:
Famous last words, but I think at this point the code coverage is the only remaining issue.

csabella:
@brettcannon   I don't understand what changes to make to fix code coverage.    I know the tests are hitting that line, especially since the line after it is green.  I'm sure it's due to my lack of knowledge on code cov, but do you have any suggestions?  Thanks!

tests/test_news.py
```diff
@@ -57,13 +57,14 @@ def test_date_only(self):
         'action': 'opened',
         'number': 1234,
         'pull_request': {
+            'url': 'https://api.github.com/repos/cpython/python/pulls/1234',
             'statuses_url': 'https://api.github.com/some/status',
             'issue_url': 'https://api.github.com/repos/cpython/python/issue/1234',
         },
     }
     event = sansio.Event(event_data, event='pull_request', delivery_id=1)
```
Mariatta: I believe event is now an unused variable.

tests/test_prtype.py
```diff
@@ -0,0 +1,205 @@
...
+    event_data = {
+        'action': 'opened',
+        'number': 1234,
+        'pull_request': {
+            'url': 'https://api.github.com/repos/cpython/python/pulls/1234',
+            'statuses_url': 'https://api.github.com/some/status',
+            'issue_url': 'https://api.github.com/repos/cpython/python/issue/1234',
+        },
+    }
+    event = sansio.Event(event_data, event='pull_request', delivery_id=1)
```
Mariatta: `event` is unused variable.

bedevere/prtype.py
```diff
@@ -0,0 +1,51 @@
...
+        filepath = pathlib.PurePath(filename)
+        if filepath.suffix == '.rst':
+            docs = True
+        elif filepath.name.startswith('test_'):
+            tests = True
+        else:
+            return
+    if tests:
+        await add_category(gh, issue, Category.tests)
+    elif docs:
```
Mariatta: Seems like there is no test if `filenames` is empty, which will lead to `tests` and `docs` to be both False.
I think codecov is expecting that path, where both `tests` and `docs` are empty.

csabella: Thanks @Mariatta .  I really thought that would fix it, but now it seems worse.   :-(  Maybe I messed it up, so I need to look at it more.

tests/test_prtype.py
```diff
@@ -0,0 +1,216 @@
...
+        self.post_url.append(url)
+        self.post_data.append(data)
+        return self._post_return
+
+
+GOOD_BASENAME = '2017-06-16-20-32-50.bpo-1234.nonce.rst'
+
+
+
+async def no_files():
```
Mariatta: Just need to prefix this with `test_`, then that should take care of coverage.

csabella: Wow, thank you so much!  That fixed it.  Apologies for not catching it myself.

brettcannon:
Thanks, @csabella !

PR #110, opened 2018-06-13
Use Python 3.6.5 in Heroku

Mariatta:


Mariatta:
I also notice that the "master" branch is not protected..

Issue #111, opened 2018-06-24
Check consistency between merge target and title.


terryjreedy:
https://github.com/python/cpython/pull/7765
is a trivial doc change intended for master, 3.7, 3.6, and 2.7.  I reviewed,
approved, marked labels, and after a 2nd approval, merged it. miss-islington
cherry-picked it to 3.7, but failed for 3.6 and 2.7.  The author then discovered
that the 3.6 failure was because the original PR was (accidentally) against
3.6, not master.

On core_mentorship, Nick suggested that the best way to catch such an error is for Bedevere to check consistency between between branch merge target (in above case, 3.6) and the presence or absence of a branch label (in above case, none). 

"When bedevere checks for the BPO issue reference, it could also check
that the target branch matches the commit message prefix (i.e. if
there's no "[3.6] " at the start, then targetting the 3.6 branch is
wrong - PRs without a version prefix should always be targeting
master, and PRs deliberately targeting a maintenance branch directly
should include the prefix).
"
Possible message: 
"If this PR is really for 3.6, prefix [3.6] to the title.
If this PR is for master, CLOSE IT and make a new PR."

Mariatta:
Duplicate of https://github.com/python/core-workflow/issues/211

PR #112, opened 2018-06-24
Post a status check for backport PR title.

Mariatta:
If a PR is made not against master, check the PR title.
If the PR title does not match the backport pattern,
post a failure status.

Two ways to resolve:
- reopen the PR against master
- Update the PR title accordingly

Depends on https://github.com/python/devguide/pull/392

- [x] Merge https://github.com/python/devguide/pull/392 first

Closes https://github.com/python/core-workflow/issues/211
Closes https://github.com/python/bedevere/issues/111

### DIFF
--- a/bedevere/backport.py
+++ b/bedevere/backport.py
@@ -69,9 +69,10 @@ async def manage_labels(event, gh, *args, **kwargs):
 @router.register("pull_request", action="edited")
 @router.register("pull_request", action="synchronize")
 async def validate_backport_pr(event, gh, *args, **kwargs):
-    """ Detect when PRs made against maintenance branch, and the title does not
-    match backport PR pattern.
-    Post a failure status check.
+    """Check the PR title for backport pull requests.
+
+    If the PR was made against maintenance branch, and the title does not
+    match the backport PR pattern, then post a failure status.
     """
     if event.data["action"] == "edited" and "title" not in event.data["changes"]:
         return
--- a/bedevere/backport.py
+++ b/bedevere/backport.py
@@ -1,9 +1,12 @@
-"""Automatically remove a backport label."""
+"""Automatically remove a backport label, and check backport PR validity."""
+import functools
 import re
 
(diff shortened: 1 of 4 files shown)


bedevere/backport.py
```diff
@@ -58,3 +63,33 @@
...
     await _copy_over_labels(gh, original_issue, backport_issue)
+
+
+@router.register("pull_request", action="opened")
+@router.register("pull_request", action="edited")
+@router.register("pull_request", action="synchronize")
+async def validate_backport_pr(event, gh, *args, **kwargs):
+    """ Detect when PRs made against maintenance branch, and the title does not
+    match backport PR pattern.
+    Post a failure status check.
```
brettcannon: Missing a newline.

bedevere/backport.py
```diff
@@ -58,3 +63,33 @@
...
+        return
+    pull_request = event.data["pull_request"]
+    base_branch = pull_request["base"]["ref"]
+
+    if base_branch == "master":
+        return
+
+    title = util.normalize_title(pull_request["title"],
+                                 pull_request["body"])
+    title_match = TITLE_RE.match(title)
```
brettcannon: Is there a reason to worry about PRs that are only to fix a bug in a maintenance branch and has no commit against `master`? I'm thinking of the case when something was implicitly fixed by something and the need to fix a maintenance release comes up later. In that instance there wouldn't necessarily be a PR to refer to.

Mariatta: I'm thinking that PR to the maintenance branch that is not a backport of another PR should be quite rare, and if happens we should be "suspicious" about it. That's why I'm checking for both [X.Y] and (GH-NNNN) here.
But I propose to make this status check optional, so we can still merge it even if the status check failed.

brettcannon: Fair enough. We can see if anyone complains about the failure being a big deal on this cases and adjust accordingly.

PR #144, opened 2018-12-08
leave codeblocks untouched while hyperlinking

### LINKED ISSUE #136: Automatically hyperlinking of `bpo-` text sometimes causes confusion
Automatically changing bpo-DDDD to a link to bugs.python.org was introduced in #121 . It's a useful feature but there are a few cases where the PR was raised with wrong title format and while specifying the correct format and the bot rewrites bpo number in the text as a link causing confusion. It does rewriting without notification and thus the comments are silently overwritten unless OP re-reads it.

Example : While commenting the correct format in https://github.com/python/cpython/pull/9906#issuecomment-430177012 bedevere changed `bpo-34996: Add name parameter to proccess and thread pool` to `[bpo-34996](https://bugs.python.org/issue34996): Add name parameter to proccess and thread pool` . The user was confused and changed the PR title with URL. Using backticks also doesn't stop the bot. Is there a way to specify bpo-DDDD without adding hyperlinks in the comments? Maybe this can be skipped in the backticks?

Another example where spaces have to be used to stop rewriting : https://github.com/python/cpython/pull/9755#issuecomment-430679326

Thanks

danishprakash:
Almost always, we would want the text inside a code block to be the way it is/untouched. This has, previously, lead to a few problems in discussions and otherwise. See the discussion over at #136.

Closes #136

### DIFF
--- a/bedevere/bpo.py
+++ b/bedevere/bpo.py
@@ -140,17 +140,22 @@ def create_hyperlink_in_comment_body(body):
     new_body = ""
     leftover_body = body
     ISSUE_RE = re.compile(r"bpo-(?P<issue>\d+)")
-
-    # leave code blocks, inline or otherwise, untouched
-    CODE_BLOCK = re.compile(r"(`{1,3})(.|\s)*(`{1,3})")
-    if CODE_BLOCK.search(body):
-        return body
+    CODE_BLOCK = re.compile(r"(`{3}([^`]*?)`{3})|(^`([^`]*?)`$)")  # leave code blocks, inline or otherwise, untouched
 
     while True:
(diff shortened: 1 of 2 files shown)


brettcannon:
@danishprakash please add tests to get code coverage back up to 100%.

nightlark:
Is this PR one that can be closed? The GitHub autolink feature enabled as part of https://github.com/python/core-workflow/issues/361 seems to handle the case of leaving codeblocks (and things enclosed by backticks) untouched.

brettcannon:
Yes, we can close it as out-of-date (sorry to @danishprakash if they did fix things as requested and people forgot to notify folks to review again).

PR #147, opened 2018-12-14
Fix broken link in README

lysnikolaou:
Updated a link in the README that was broken(pointing to the old devguide)


brettcannon:
Thanks!

PR #152, opened 2019-02-11
Check news will pass as long as the file is not empty.

Mariatta:
From discussions in https://github.com/python/cpython/pull/11779#issuecomment-462498706, it was determined that the check news status should pass as long as the news file is not empty. cc @ned-deily


ned-deily:
Thank you!

PR #154, opened 2019-02-12
Use the xenial dist when testing Python 3.7

Mariatta:
Ref: https://github.com/travis-ci/travis-ci/issues/9069

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,11 @@
 language: python
 cache: pip
 python:
-  - 3.6-dev
+  - 3.6
+  - 3.7
   - 3.7-dev
   - nightly
-
-matrix:
-  include:
-    - python: "3.7"
-      dist: xenial
+dist: xenial
 
 before_install:
   - python3 -m pip install coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - 3.6
   - 3.7
   - 3.7-dev
+  - 3.8-dev
   - nightly
 dist: xenial
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ python:
   - 3.7-dev
   - nightly
 
+matrix:
+  include:
+    - python: "3.7"
+      dist: xenial
+
 before_install:
   - python3 -m pip install coverage
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,13 @@ python:
   - nightly
 dist: xenial
 
+matrix:
+  allow_failures:
+  - python: 3.8-dev
+    dist: xenial
+  - python: nightly
+    dist: xenial
+
 before_install:
   - python3 -m pip install coverage
 install:
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,5 +9,5 @@ packaging==19.0
 py==1.7.0
 pyparsing==2.3.1
 six==1.12.0
-uritemplate==3.0.0
+-e git://github.com/python-hyper/uritemplate.git@68064e2e8f60687674071f9c791603b0ae6a76a4#egg=uritemplate
 yarl==1.3.0


Mariatta:
Reading this (https://docs.travis-ci.com/user/languages/python/#development-releases-support), trusty has outdated Python versions. Python3.7-dev in there was really Python 3.7.0a4 instead of the latest release.

With xenial, `3.7` gives us Python 3.7.1, and `3.7-dev` gives us Python 3.7.2+.

Mariatta:
Seems like there are other failing tests related to aiohttp in Python 3.8.0a1:

```
______________________________ test_ping[pyloop] _______________________________
self = <aiohttp.client_reqrep.ClientRequest object at 0x7f4c311a0b80>
writer = <aiohttp.http_writer.StreamWriter object at 0x7f4c311a00a0>
conn = Connection<ConnectionKey(host='127.0.0.1', port=38033, is_ssl=False, ssl=None, proxy=None, proxy_auth=None, proxy_headers_hash=None)>
    async def write_bytes(self, writer: AbstractStreamWriter,
                          conn: 'Connection') -> None:
        """Support coroutines that yields bytes objects."""
        # 100 response
        if self._continue is not None:
            await writer.drain()
            await self._continue
    
        protocol = conn.protocol
        assert protocol is not None
        try:
            if isinstance(self.body, payload.Payload):
>               await self.body.write(writer)
../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/aiohttp/client_reqrep.py:553: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <aiohttp.payload.JsonPayload object at 0x7f4c313033d0>
writer = <aiohttp.http_writer.StreamWriter object at 0x7f4c311a00a0>
    async def write(self, writer: AbstractStreamWriter) -> None:
>       await writer.write(self._value)
../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/aiohttp/payload.py:231: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <aiohttp.http_writer.StreamWriter object at 0x7f4c311a00a0>
chunk = b'{"zen": "testing is good"}'
    async def write(self, chunk: bytes,
                    *, drain: bool=True, LIMIT: int=0x10000) -> None:
        """Writes chunk of data to a stream.
    
        write_eof() indicates end of stream.
        writer can't be used after write_eof() method being called.
        write() return drain future.
        """
        if self._on_chunk_sent is not None:
            await self._on_chunk_sent(chunk)
    
        if self._compress is not None:
            chunk = self._compress.compress(chunk)
            if not chunk:
                return
    
        if self.length is not None:
            chunk_len = len(chunk)
            if self.length >= chunk_len:
                self.length = self.length - chunk_len
            else:
                chunk = chunk[:self.length]
                self.length = 0
                if not chunk:
                    return
    
        if chunk:
            if self.chunked:
                chunk_len_pre = ('%x\r\n' % len(chunk)).encode('ascii')
                chunk = chunk_len_pre + chunk + b'\r\n'
    
>           self._write(chunk)
../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/aiohttp/http_writer.py:101: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <aiohttp.http_writer.StreamWriter object at 0x7f4c311a00a0>
chunk = b'{"zen": "testing is good"}'
    def _write(self, chunk: bytes) -> None:
        size = len(chunk)
        self.buffer_size += size
        self.output_size += size
    
        if self._transport is None or self._transport.is_closing():
>           raise ConnectionResetError('Cannot write to closing transport')
E           ConnectionResetError: Cannot write to closing transport
../../../virtualenv/python3.8-dev/lib/python3.8/site-packages/aiohttp/http_writer.py:67: ConnectionResetError
The above exception was the direct cause of the following exception:
aiohttp_client = <function aiohttp_client.<locals>.go at 0x7f4c3124ad40>
    async def test_ping(aiohttp_client):
        app = web.Application()
        app.router.add_post("/", main.main)
        client = await aiohttp_client(app)
        headers = {"x-github-event": "ping",
                   "x-github-delivery": "1234"}
        data = {"zen": "testing is good"}
>       response = await client.post("/", headers=headers, json=data)
tests/test___main__.py:14: 
```

Mariatta:
The issue with uritemplate is fixed if we're using the latest changes (not yet released).
But now there is incompatibility with Python 3.8 but I think it is outside of comfort zone.

.travis.yml
```diff
@@ -1,9 +1,12 @@
 language: python
 cache: pip
 python:
-  - 3.6-dev
+  - 3.6
+  - 3.7
   - 3.7-dev
+  - 3.8-dev
```
brettcannon: Isn't there a way to make testing against 3.8 an allowed failure so it's more informational than anything?

Mariatta: Thanks. I've made it so failures allowed in `3.8-dev` and `nightly`,

Mariatta:
I'll also have to keep an eye for when uritemplate > 3.0.0 gets released 🤨

PR #159, opened 2019-03-23
Validate bpo issue number

lysnikolaou:
Because #72 was closed I thought I could move the work of Lev forward. Taking some of his work and following all the suggestions in the initial PR, I have formulated a new one.

I used the `requests` library, but did not updated the `requirements.txt` file, because I'm not sure that the correct way to go. `urllib` might be a better alternative, because it's in the stdlib.

Relevant issue: #15 
Co-authored-by: Lev Lazinskiy

Update: I updated `requirements.txt` so that the tests can pass on Travis. In case we use some other library, it can always be updated again.


brettcannon:
Just an FYI I'm not comfortable accepting this with the use of `requests` as that's synchronous and the design of the bot is purposefully async. Please use `aiohttp` for any HTTP calls.

lysnikolaou:
Sorry @brettcannon, hadn't thought of that. I updated my PR to use `aiohttp`.

bedevere/bpo.py
```diff
@@ -49,7 +44,12 @@
                 new_body = BODY.format(body=body, issue_number=issue_number)
                 body_data = {"body": new_body, "maintainer_can_modify": True}
                 await gh.patch(event.data["pull_request"]["url"], data=body_data)
-        status = create_success_status(issue_number_found)
+        issue_number_on_bpo = (await
```
brettcannon: I wouldn't split this line. Just assign `issue_number_found.group()` to a variable and then pass that into the `await` call. That will probably also shorten it enough to be able to put it in the first line of the `if`.

bedevere/bpo.py
```diff
@@ -1,5 +1,6 @@
 """Check if a bugs.python.org issue number is specified in the pull request's title."""
 import re
+import aiohttp
```
brettcannon:
```suggestion

import aiohttp
```

bedevere/bpo.py
```diff
@@ -111,6 +111,20 @@ def create_success_status(found_issue):
                               target_url=url)
 
 
+def create_failure_status(issue_not_found_on_bpo, found_issue=None):
```
brettcannon: There's no need to overload creating a failure case for the missing issue number in the PR versus the issue not existing. Just make them separate functions.

bedevere/bpo.py
```diff
@@ -154,3 +168,11 @@ def create_hyperlink_in_comment_body(body):
             leftover_body = leftover_body[presence:]
     new_body = new_body + leftover_body
     return new_body
+
+
+async def _validate_issue_number(issue_number):
+    url = f"https://bugs.python.org/issue{issue_number}"
+    async with aiohttp.ClientSession() as session:
+        async with session.head(url) as res:
+            return False if res.status == 404 else True
```
brettcannon:
```suggestion
            return res.status != 404
```

tests/test_bpo.py
```diff
@@ -50,6 +50,32 @@ def __init__(self, *, getitem=None):
     assert status["context"] == "bedevere/issue-number"
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("action", ["opened", "synchronize", "reopened"])
+async def test_set_status_failure_via_issue_non_found_on_bpo(action):
```
brettcannon:
```suggestion
async def test_set_status_failure_via_issue_not_found_on_bpo(action):
```

bedevere/bpo.py
lysnikolaou: Are you referring to lines 37-38 or 47-48?

bedevere/bpo.py
lysnikolaou: Haha, that's kind of embarrassing...

bedevere/bpo.py
lysnikolaou: Should there be an empty line between `import aiohttp` and `from gidgethub import routing`?

lysnikolaou:
Thanks for the review @brettcannon, I have updated my PR.

bedevere/bpo.py
brettcannon: This hasn't been fixed yet.

lysnikolaou: There is now a blank line between `import re` and `import aiohttp`. Should there be a blank line between `import aiohttp` and `from gidgethub import routing`?

brettcannon: Please follow https://www.python.org/dev/peps/pep-0008/#imports.

lysnikolaou: I'm terribly sorry, but I can't see how the updated version does not follow PEP8. On the GitHub 'Files changed' interface there is indeed a blank line between `stdlib` and third parties `import`s, which is what PEP8 specifies. It is also consistent with how `__main__.py` does `imports, so I'm not exactly sure what I should change.

brettcannon: So sorry, you're right. I've received a flood or PRs this week and I was thinking about the gidgethub repo. 😄

lysnikolaou:
@brettcannon Thanks again for the review. I have now pushed one more commit which adds the necessary docstrings, so the PR might be ready to be merged.

brettcannon:
I don't quite have time to review again, but I did notice your docstrings weren't complete sentences (i.e. missing periods). We follow PEP 8 as a style so please update the docstrings (I know this can seem very nit-picky, but it helps readers know you actually meant for the sentence to end and didn't accidentally leave something out).

lysnikolaou:
I added the missing periods, @brettcannon. Please let me know in case there is any way I can help out in making the PR easier to review.

brettcannon:
@lysnikolaou unless you can magically find me more time, it's just a matter of work easing up a bit (I'm hoping I will get to this either this week or next).

bedevere/bpo.py
```diff
@@ -40,16 +34,21 @@
...
+                                    else create_failure_status_no_issue())
     else:
+        issue_number = issue_number_found.group("issue")
         if "body" in event.data["pull_request"]:
             body = event.data["pull_request"]["body"] or ""
             if not body or CLOSING_TAG not in body:
-                issue_number = issue_number_found.group("issue")
                 new_body = BODY.format(body=body, issue_number=issue_number)
                 body_data = {"body": new_body, "maintainer_can_modify": True}
                 await gh.patch(event.data["pull_request"]["url"], data=body_data)
```
brettcannon: The patching of the first message seems to occur regardless of whether the issue # was validated. Do you think it would be better to not create a URL if we can check that it won't return a 404?

bedevere/bpo.py
```diff
@@ -40,16 +34,21 @@
...
-        status = create_success_status(issue_number_found)
+        issue_number_on_bpo = await _validate_issue_number(issue_number)
+        if issue_number_on_bpo:
+            status = create_success_status(issue_number_found)
```
brettcannon: Any reason to not pass in `issue_number` since you already have that pulled out?

bedevere/bpo.py
```diff
@@ -111,6 +110,25 @@ def create_success_status(found_issue):
                               target_url=url)
 
 
+def create_failure_status_issue_not_on_bpo(issue_number_found):
+    """Create a failure status for when an issue does not exist on the bug tracker."""
+    issue_number = issue_number_found.group("issue")
```
brettcannon: What is the benefit of not passing in the issue number directly instead of the `re.Match` object?

bedevere/bpo.py
```diff
@@ -154,3 +172,12 @@ def create_hyperlink_in_comment_body(body):
...
     new_body = new_body + leftover_body
     return new_body
+
+
+async def _validate_issue_number(issue_number):
+    """Make sure the issue number exists on bugs.python.org."""
+    url = f"https://bugs.python.org/issue{issue_number}"
+    async with aiohttp.ClientSession() as session:
+        async with session.head(url) as res:
+            return res.status != 404
```
brettcannon: Isn't it more true that as long as it isn't a 40x status? Or is there no way to get anything but a 404 back?

lysnikolaou: There are other 40x status codes that correspond to a client error, that does not have to with the page not being found, i.e. `408 Request Timeout`. If bpo follows these conventions, then 404 should be the only possible status code to receive in the response.

bedevere/bpo.py
lysnikolaou: When I created the two functions `create_failure_status_issue_not_on_bpo` and `create_failure_status_no_issue`, I wanted them to be consistent with how `create_success_status` was already doing things. That's why I passed `issue_number_found` as a parameter and not `issue_number`. I agree that it would be better to pass `issue_number` though, since that is already there.

bedevere/bpo.py
lysnikolaou: See above.

bedevere/bpo.py
lysnikolaou: Yeah, that sounds like a good improvement.

lysnikolaou:
@brettcannon we should maybe also consider that validating issues in the tests has really slowed down `test_bpo`, because it has to wait for the responses. Do you think that's an issue or is it okay?

brettcannon:
@lysnikolaou Stubbing/mocking/patching out the b.p.o calls for all tests but the one(s) that are specifically testing this functionality would be good.

When you're ready for me to re-review, let me know (I think you can click on my review status to ask for a new review now in GitHub).

lysnikolaou:
@brettcannon Sorry for my delayed response, I had no time at all in the few last days. I am not sure, how I could do that, since there is no direct calling of the functions from the tests. Should we maybe add another field to the `event` parameter and the check for that in `set_status`? That doesn't seem like the best way to do it, so a little bit of guidance would be needed.

brettcannon:
@lysnikolaou you can use https://docs.pytest.org/en/latest/monkeypatch.html. And don't worry about the test mocking out functions that are not publicly exposed as we're fine with glass box testing.

lysnikolaou:
@brettcannon I have patched out the bpo calls from all the tests that would make it and do not test this specifically. I am not sure, if that's the best way to do it, so your review is needed once more.

bedevere/bpo.py
```diff
@@ -40,16 +34,24 @@
...
-            if not body or CLOSING_TAG not in body:
-                issue_number = issue_number_found.group("issue")
-                new_body = BODY.format(body=body, issue_number=issue_number)
-                body_data = {"body": new_body}
-                await gh.patch(event.data["pull_request"]["url"], data=body_data)
-        status = create_success_status(issue_number_found)
+        issue_number = issue_number_found.group("issue")
+        try:
+            issue_number_on_bpo = await _validate_issue_number(issue_number)
+        except NameError:
```
brettcannon: What is triggering a `NameError`? That shouldn't be occurring and so that makes this `try` block suspicious. If it's because of how `aiohttp` functions then it should be contained within `_validate_issue_number()` so it effectively can't fail unless there's a critical exception that should bubble up to get logged.

bedevere/bpo.py
```diff
@@ -40,16 +34,24 @@
...
+            issue_number_on_bpo = True
+        if issue_number_on_bpo:
+            if "body" in event.data["pull_request"]:
+                body = event.data["pull_request"]["body"] or ""
+                if not body or CLOSING_TAG not in body:
+                    new_body = BODY.format(body=body, issue_number=issue_number)
+                    body_data = {"body": new_body, "maintainer_can_modify": True}
```
brettcannon: Is `"maintainer_can_modify"` required for this to work, or are you just sneaking that in here? If it's the latter then I would pull it out as it's easy to lose track of. Better to open an issue and do a separate check for it as this isn't related to issue numbers from bugs.python.org which is what this module is aimed at.

tests/test_bpo.py
```diff
@@ -52,7 +53,34 @@ def __init__(self, *, getitem=None):
...
+    assert status["state"] == "failure"
+    assert status["target_url"].startswith("https://devguide.python.org")
+    assert status["context"] == "bedevere/issue-number"
+    assert status["description"] == "Issue #123 not found on bugs.python.org"
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("action", ["opened", "synchronize", "reopened"])
+async def test_set_status_success(action, monkeypatch):
+    monkeypatch.delattr('bedevere.bpo._validate_issue_number')
```
brettcannon: The problem with the deletion of the function is now you don't know if the function was called with the appropriate issue number. I would set up a mock and verify that it was called with the expected issue number. That way you can also return false as expected to make sure everything is calling that function the way you expect it to.

bedevere/bpo.py
```diff
@@ -154,3 +173,12 @@ def create_hyperlink_in_comment_body(body):
             leftover_body = leftover_body[presence:]
     new_body = new_body + leftover_body
     return new_body
+
+
+async def _validate_issue_number(issue_number):
+    """Make sure the issue number exists on bugs.python.org."""
+    url = f"https://bugs.python.org/issue{issue_number}"
+    async with aiohttp.ClientSession() as session:
```
brettcannon: Instead of creating a new session for `aiohttp`, you could pass it in through the `__main__` method where dispatch occurs to try and maximize connection pooling. All routes should be acceptings `*args, **kwargs` so passing in a `session` keyword argument should work fine.

lysnikolaou:
Hey @brettcannon, sorry for the delay again. Life gets in the way. For the last few days, I've been trying to implement all the proposed changes and I have an implementation that uses `pytest-mock`. First of all would that be ok? Second of all, if I have understood correctly, one cannot mock an `async` function. I have found the [asynctest](https://pypi.org/project/asynctest/) library, which provides a `CoroutineMock`. Should I use that or is there maybe another, easier way?

brettcannon:
It isn't like I'm very fast on my reviews so no need to apologize on taking your time. 😄 

As for mocking, I have no specific guidance beyond you can use the hand-coded mocks as necessary, but otherwise I have no opinion on what 3rd-party packages to use.

lysnikolaou:
@brettcannon I used `asynctest` to mock out `_validate_issue_number` and it works alright up to a single method, which is successful on my local machine and when I debug it. Would you have an idea, what's wrong here?

Moreover, sending a kwarg in `__main__.py` does not work for me? Should it just work out of the box, without any additional considerations?

brettcannon:
@lysnikolaou I think the mocking problem is that you are using a fixture to blindly assign the mock, which means you never actually put the object you mocked out back. I would drop your custom fixture and instead use http://doc.pytest.org/en/latest/monkeypatch.html and set a mock as needed in each test. That way `monkeypatch` can clean up the mock afterwards and put everything back to the way it was.

As for your `__main__.py` question, I realize my comment was way too vague. 😄  What I meant to say if you could tweak the following to pass in `session=session`:

https://github.com/python/bedevere/blob/master/bedevere/__main__.py#L40

lysnikolaou:
@brettcannon Both monkeypatching and the passing of session from `main` are now implemented. Because `_validate_issue_number` is being monkeypatched out during most of the tests, I do not create new session objects for those, which means that `session` gets the default value `""` during those and then that gets passed to the mocked function. Do you think that's okay like that?

bedevere/bpo.py
```diff
@@ -40,16 +33,22 @@
...
-        if "body" in event.data["pull_request"]:
-            body = event.data["pull_request"]["body"] or ""
-            if not body or CLOSING_TAG not in body:
-                issue_number = issue_number_found.group("issue")
-                new_body = BODY.format(body=body, issue_number=issue_number)
-                body_data = {"body": new_body}
-                await gh.patch(event.data["pull_request"]["url"], data=body_data)
-        status = create_success_status(issue_number_found)
+        issue_number = issue_number_found.group("issue")
+        session = kwargs.get('session', '')
```
brettcannon: Instead of doing this I would add `session` as keyword argument to this function with a default value of `None`.

bedevere/bpo.py
```diff
@@ -104,15 +103,32 @@
...
     url = f"https://bugs.python.org/issue{issue_number}"
     return util.create_status(STATUS_CONTEXT, util.StatusState.SUCCESS,
                               description=f"Issue number {issue_number} found",
                               target_url=url)
 
 
+def create_failure_status_issue_not_on_bpo(issue_number):
+    """Create a failure status for when an issue does not exist on the bug tracker."""
+    description = f"Issue #{issue_number} not found on bugs.python.org"
+    url = "https://devguide.python.org/pullrequest/#submitting"
```
brettcannon: What do you think of changing this to be the issue URL to show that it's a 404? If they specified an issue number they know that's what they should be doing so I don't think linking back to the docs is going to be the most helpful option for us.

tests/test_bpo.py
```diff
@@ -48,11 +52,36 @@ def __init__(self, *, getitem=None):
...
+    data = {
+        "action": action,
+        "pull_request": {
+             "statuses_url": "https://api.github.com/blah/blah/git-sha",
+            "title": "bpo-123: Invalid issue number",
+        },
+    }
+    event = sansio.Event(data, event="pull_request", delivery_id="12345")
+    gh = FakeGH()
+    session = aiohttp.ClientSession()
```
brettcannon: Any reason not to use the context manager to make sure it closes?

tests/test_bpo.py
```diff
@@ -405,6 +498,10 @@ def __init__(self, *, getitem=None):
     assert body_data["body"].count("[bpo-123](https://bugs.python.org/issue123)") == 2
     assert body_data["body"].count("[something about bpo-123](https://bugs.python.org/issue123)") == 1
     assert "123456" in gh.patch_url
+    if action == 'opened':
```
brettcannon: If this is necessary then it might be better to have a helper function for the tests which does everything that's the same and then break this into two separate tests that do the appropriate mock check that's unique to each.

bedevere/bpo.py
lysnikolaou: I totally agree. I will change the url to link to the (non-existing) issue.

bedevere/bpo.py
lysnikolaou: Of course!

tests/test_bpo.py
```diff
@@ -69,11 +97,36 @@ def __init__(self, *, getitem=None):
     assert "1234" in status["description"]
     assert status["context"] == "bedevere/issue-number"
     assert "git-sha" in gh.url
+    bpo._validate_issue_number.assert_awaited_with("1234", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("1234", session=None)
```

tests/test_bpo.py
```diff
@@ -110,10 +166,13 @@ def __init__(self, *, getitem=None):
     gh = FakeGH()
     await bpo.router.dispatch(event, gh)
     assert gh.data is not None
+    bpo._validate_issue_number.assert_awaited_with("1234", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("1234", session=None)
```

tests/test_bpo.py
```diff
@@ -131,10 +190,13 @@ def __init__(self, *, getitem=None):
     await bpo.router.dispatch(event, gh)
     assert gh.patch_data is not None
     assert gh.patch_data["body"] == "\n\n<!-- issue-number: bpo-32636 -->\nhttps://bugs.python.org/issue32636\n<!-- /issue-number -->\n"
+    bpo._validate_issue_number.assert_awaited_with("32636", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("32636", session=None)
```

tests/test_bpo.py
```diff
@@ -239,10 +307,13 @@ def __init__(self, *, getitem=None):
     assert "1234" in status["description"]
     assert status["context"] == "bedevere/issue-number"
     assert "git-sha" in gh.url
+    bpo._validate_issue_number.assert_awaited_with("1234", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("1234", session=None)
```

tests/test_bpo.py
```diff
@@ -273,10 +347,13 @@ def __init__(self, *, getitem=None):
     status = gh.patch_data
     assert "https://bugs.python.org/issue1234" in status["body"]
     assert "1347" in gh.patch_url
+    bpo._validate_issue_number.assert_awaited_with("1234", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("1234", session=None)
```

tests/test_bpo.py
```diff
@@ -291,11 +368,24 @@ def __init__(self, *, getitem=None):
     await bpo.router.dispatch(event, gh)
     assert gh.patch_data is None
     assert gh.patch_url is None
+    bpo._validate_issue_number.assert_awaited_with("1234", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("1234", session=None)
```

tests/test_bpo.py
```diff
@@ -291,11 +368,24 @@ def __init__(self, *, getitem=None):
...
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("action", ["opened", "edited"])
-async def test_set_pull_request_body_success(action):
+async def test_set_pull_request_body_success_opened(monkeypatch):
+    await set_pull_request_body_success_helper("opened", monkeypatch)
+    bpo._validate_issue_number.assert_awaited_with("12345", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("12345", session=None)
```

tests/test_bpo.py
```diff
@@ -356,6 +450,7 @@ def __init__(self, *, getitem=None):
     await bpo.router.dispatch(event, gh)
     if gh.patch_data:
         assert "[bpo-123](https://bugs.python.org/issue123)" not in gh.patch_data
+        bpo._validate_issue_number.assert_awaited_with("12345", None)
```
brettcannon:
```suggestion
        bpo._validate_issue_number.assert_awaited_with("12345", session=None)
```

tests/test_bpo.py
```diff
@@ -380,8 +475,26 @@ def __init__(self, *, getitem=None):
...
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("action", ["opened", "edited"])
-async def test_set_pull_request_body_already_hyperlinked_bpo(action):
+async def test_set_pull_request_body_already_hyperlinked_bpo_opened(monkeypatch):
+    await set_pull_request_body_already_hyperlinked_bpo_helper(
+        "opened",
+        monkeypatch
+    )
+    bpo._validate_issue_number.assert_awaited_with("12345", None)
```
brettcannon:
```suggestion
    bpo._validate_issue_number.assert_awaited_with("12345", session=None)
```

lysnikolaou:
@brettcannon Thanks for all the reviews and having such great patience.

Issue #165, opened 2019-04-23
Automatically request new review when state changed to "awaiting change review"


brettcannon:
Now that GitHub has the concept of a "review request", we could take advantage of that when Bedevere is triggered to message people requesting changes by requesting reviews again from those people. That way we get both the label to tell people when viewing their https://github.com/pulls/review-requested page, but also any other part of the GitHub UI which renders review states.

lysnikolaou:
Would it maybe make sense to wait until the warning on the [Review Requests API page](https://developer.github.com/v3/pulls/review_requests/) gets out of the way?

> Warning: The API may change without advance notice during the preview period. Preview features are not supported for production use. If you experience any issues, contact GitHub Support.

brettcannon:
@lysnikolaou it depends on whether that warning applies to just the emoji support or the API all-up. Plus I didn't want to forget the idea so I opened this issue regardless.

lysnikolaou:
You are right, @brettcannon. After some digging, I found out that the API itself is **not** in beta, so it must be ready to use in production. I'll try to have a PR ready in the next days.

brettcannon:
@lysnikolaou you're right. :)

PR #168, opened 2019-05-13
Request review on GitHub when the PR needs change review

lysnikolaou:
Upon changing the status to `awaiting change review`, a review is also requested through the new GitHub API from the appropriate reviewers.

#165

### DIFF
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -222,6 +222,7 @@ async def new_comment(event, gh, *args, **kwargs):
             thanks = BORING_THANKS
         comment = ACK.format(greeting=thanks, core_devs=core_devs)
         await gh.post(issue["comments_url"], data={"body": comment})
+        # Re-request reviews from core developers based on the new state of the PR.
         reviewers_url = pr_url + '/requested_reviewers'
         reviewers = [core_dev async for core_dev in core_dev_reviewers(gh, pr_url)]
         await gh.post(reviewers_url, data={"reviewers": reviewers})
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -24,7 +24,7 @@ digraph "PR stages" {
   "Awaiting review" -> "Awaiting core review" [label="New review", color=blue]
   "Awaiting core review" -> "Awaiting core review" [label="New review", color=blue]
   "Awaiting core review" -> "Awaiting changes" [label="New review requests changes", color=green]
-  "Awaiting changes" -> "Awaiting change review" [label="Comments changes are done", color=orange]
+  "Awaiting changes" -> "Awaiting change review" [label="Comments changes are done\nBedevere requests review from core-dev", color=orange]
   "Awaiting change review" -> "Awaiting changes" [label="New review requests changes", color=green]
   "Awaiting change review" -> "Awaiting merge" [label="New review approves", color=green]
 
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -222,6 +222,9 @@ async def new_comment(event, gh, *args, **kwargs):
             thanks = BORING_THANKS
         comment = ACK.format(greeting=thanks, core_devs=core_devs)
         await gh.post(issue["comments_url"], data={"body": comment})
+        reviewers_url = pr_url + '/requested_reviewers'
+        reviewers = [core_dev async for core_dev in core_dev_reviewers(gh, pr_url)]
+        await gh.post(reviewers_url, data={"reviewers": reviewers})
 
 
 @router.register("pull_request", action="closed")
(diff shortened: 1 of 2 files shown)


bedevere/stage.py
```diff
@@ -222,6 +222,9 @@ class Blocker(enum.Enum):
             thanks = BORING_THANKS
         comment = ACK.format(greeting=thanks, core_devs=core_devs)
         await gh.post(issue["comments_url"], data={"body": comment})
+        reviewers_url = pr_url + '/requested_reviewers'
```
brettcannon: Too bad that isn't in the webhook data somewhere. 😞

brettcannon: Maybe a comment on what this is doing/for?
```suggestion
        # Re-request reviews from core developers based on the new state of the PR.
        reviewers_url = pr_url + '/requested_reviewers'
```

bedevere/stage.py
```diff
@@ -222,6 +222,9 @@ class Blocker(enum.Enum):
...
+        reviewers = [core_dev async for core_dev in core_dev_reviewers(gh, pr_url)]
+        await gh.post(reviewers_url, data={"reviewers": reviewers})
```
brettcannon: Does `"team_reviewers": []` need to be provided in the `data`?

lysnikolaou: I have not live-tested this, but my experience with the Github API Docs is that it is explicitly stated, when a JSON field is required, which is not the case here. [An example](https://developer.github.com/v3/repos/#input) is the `name` field of the "create repository" API.

bedevere/stage.py
lysnikolaou: Would it be better to extract `"/request_reviewers"` into a global variable?

brettcannon: Eh, probably not. This was just a comment of me lamenting about this.

bedevere/stage.py
brettcannon: Would you mind live-testing this in a test repo to make sure? I'm basically trying to avoid Bedevere falling over and having to do an emergency PR. 😄 I'm happy to add a needless review somewhere to help give you data to play with.

Mariatta:
Thanks @lysnikolaou for working on this. Would you be able to update the state machine at the top of stage.py to reflect the updated workflow?
I've just remembered it as I try to add "triagers" into the mix.

bedevere/stage.py
Mariatta: I confirmed that `"team_reviewers"` can be omitted.

I tested in real-life with the following snippet:
```
import asyncio
import aiohttp
from gidgethub.aiohttp import GitHubAPI
import gidgethub
import os
async def main():
    async with aiohttp.ClientSession() as session:
        gh = GitHubAPI(session, "miss-islington", oauth_token=os.getenv("GH_AUTH"))
        try:
            result = await gh.post('/repos/python/miss-islington/pulls/253/requested_reviewers',
                                   data={"reviewers": ["miss-islington"]})
        except gidgethub.BadRequest as exc:
            print(exc)
asyncio.run(main())
```

Result: https://github.com/python/miss-islington/pull/253#event-2477015012

lysnikolaou:
@Mariatta I guess I could do that, but I don't understand what you mean by "updated workflow"? If you mean the new GitHub Triagers, then I don't know what the new state should be.

bedevere/stage.py
lysnikolaou: @brettcannon Is the live-testing by Marietta enough, or should I deploy my fork and test it on a test repository?

brettcannon:
@lysnikolaou I think she means what you're changing with this PR.

lysnikolaou:
@brettcannon @Mariatta The thing is that the state machine should fundamentally remain the same. Maybe we could change the label on the edge `Awaiting Changes -> Awaiting change review` to include something like "New review gets requested", but I don't see what other changes need to be made.

brettcannon:
@lysnikolaou yep, probably just the edge label to denote the side-effect that occurs.

bedevere/stage.py
brettcannon: @lysnikolaou I trust @Mariatta so it's enough 😄

lysnikolaou:
@brettcannon I have updated the edge label, that needed tweaking, so the PR might be ready for another review.

bedevere/stage.py
```diff
@@ -222,6 +222,10 @@ class Blocker(enum.Enum):
             thanks = BORING_THANKS
         comment = ACK.format(greeting=thanks, core_devs=core_devs)
         await gh.post(issue["comments_url"], data={"body": comment})
+        # Re-request reviews from core developers based on the new state of the PR.
+        reviewers_url = pr_url + '/requested_reviewers'
```
brettcannon:
```suggestion
        reviewers_url = f'{pr_url}/requested_reviewers'
```

brettcannon:
LGTM! @Mariatta did you want to do one final review are you okay with merging this?

Mariatta:
Sorry I don't have bandwidth to review this at the moment, don't want to block the progress.

lysnikolaou:
@brettcannon Can this be merged as well?

brettcannon:
Thanks for doing this, @lysnikolaou !

PR #158, opened 2019-03-20
Apply 'awaiting core review' label to new core-dev PRs

lysnikolaou:
Related to issue #157.

PR #182, opened 2019-07-28
Fix PR state machine graph

potomak:
After #158 has been merged, the state machine graph was out of date.
This update fixes the transition from "New PR" for core developers to
"Awaiting core review" instead of "Awaiting merge".

### Before

graphviz

### After

graphviz (1)


brettcannon:
Thanks! BTW if you have any interest in improving that graph with more descriptive edges -- e.g. "new PR" is a bit ambiguous coming out of "New PR" -- then that would be greatly appreciated!

Issue #157, opened 2019-03-20
Label new PRs from core devs with `needs core review`


brettcannon:
See https://discuss.python.org/t/poll-what-label-should-prs-from-core-devs-automatically-receive/970

Issue #139, opened 2018-10-30
Non-core review downgrades core dev PRs


pganssle:
[This PR](https://github.com/python/cpython/pull/10238) from @vstinner was initially labeled "awaiting merge", which I suppose is the default behavior when a core dev makes a PR, even before review.

I (not a core dev) reviewed the code and approved the changes, which triggered bedevere to remove "awaiting merge" and **add** "awaiting core review".

I'm guessing this is because my review triggered the logic that normally transitions from "awaiting review" to "awaiting core review", and this probably checked to see if there were any reviews from core devs (there were not) and assumed that core review was necessary.

I would think the correct behavior for an *accepted* review would be to leave "awaiting merge" in place. For a *rejected* review, adding "awaiting core review" might be the right thing to do.

pganssle:
I believe the problem is [here](https://github.com/python/bedevere/blob/b295fc6297189871a0ad69608ff818cd6d28634d/bedevere/stage.py#L172).

Mariatta:
I'm thinking it's ok for a core dev's PR to be "awaiting core review" after a review by a non core-dev.
Or perhaps only add "awaiting core review" if the review is "Changes requested"?

Mariatta:
IMO Option 1) will bring the most disruptive change to our existing workflow. While it is true for core devs the "awaiting review" label is mainly for informational purpose, and does not block merging, our other bot miss-islington depend on the "awaiting merge" label for automerge. So if core devs PRs will always start with "awaiting review" now, it means most of them can't be easily automatically merged.

I would choose to go with option 2)

terryjreedy:
OK.  I have never used auomerge, so I did not consider it. 2 is with me too.

lysnikolaou:
I think, that since #157 this is no longer an issue, so it can be closed.

brettcannon:
@lysnikolaou there might still be an issue of accidentally lowering the status, so I'm not quite ready to close this issue until we have a test to make sure you can't accidentally change the state.

potomak:
> I'm not quite ready to close this issue until we have a test to make sure you can't accidentally change the state.

@brettcannon do you think we can close this issue now that there's a test specifically for this case?

brettcannon:
@potomak yes we can!

PR #183, opened 2019-07-28
Add test case (non-core dev review)

potomak:
This is a specific test case for #139.

This should be the enough to consider the issue resolved by #157.


brettcannon:
Thanks!

PR #185, opened 2019-08-13
Don't allow failure anymore in 3.8-dev

Mariatta:


Mariatta:
The asynctest library is not compatible in Python 3.8. It was recently added in #159 by @lysnikolaou 
Is there any workaround for it? Can we make do without asynctest?

lysnikolaou:
@Mariatta the asynctest library provides mocks for async functions. We could use the newly added `AsyncMock`, but we would need to update to 3.8 immediately for that.

Mariatta:
ok thanks for the info. Hmm there is not yet 3.8 Python environment in Heroku, so we can't do it 😟

Mariatta:
Seems like Python 3.8.0 is now available in Heroku. Think we can upgrade?

lysnikolaou:
I'm not sure, what would be best to do. In case @brettcannon says we should upgrade, I could go ahead and push a PR to remove the `asynctest` dependency and update all the relevant tests.

brettcannon:
I always say you should dogfood what you produce, and since this is for Python then we should use Python 3.8 if it's available.

lysnikolaou:
It seems like there is a problem with `aiohttp` as well. After I removed the `asynctest` dependency, I get `DeprecationWarning`s due to `aiohttp` calling `asyncio.shield()` with the loop argument which is deprecated as of 3.8.

Running `pytest -p no:warnings` is ok, which shows that `asynctest` can be removed without a problem.

lysnikolaou:
Opened #211 for this, which fails on Travis with 3.7, but does not with 3.8-dev.

Mariatta:
Superseded by https://github.com/python/bedevere/pull/309

PR #211, opened 2020-01-15
Remove asynctest dependency and filter out deprecation warnings

lysnikolaou:


brettcannon:
Do note that `unittest.mock.AsyncMock` is a 3.8 thing, so Travis is unhappy.

Personally I would be fine with dropping 3.7 support as part of this PR.

lysnikolaou:
@brettcannon @Mariatta Could it be that dropping 3.7 on Travis decreases the test coverage?

brettcannon:
@lysnikolaou it's possible. Won't know without looking at what lines suddenly had their coverage change. It could also be the latest release of coverage.py 5 changing its results.

Mariatta:
I tried pulling the PR locally and ran the tests and coverage.
I tried using both coverage 5.0.3 and 4.5.4, and the result between the two are consistent, e.g. they complain about the same lines not having enough coverage. 😕Not sure why some of these weren't a problem when we were using Python 3.7.

I think we should work on improving the coverage. I might be able to look into this more in the weekend.

Thanks.

brettcannon:
The runtime file should also get updated as part of this PR if we drop 3.7 support (see #215 ).

brettcannon:
Thanks, @lysnikolaou !

PR #407, opened 2022-02-22
Improve `ci.yml`

sobolevn:
Changes:
1. Updates `codecov` version to the latest one
2. Fixes cache key, it should include both `txt` files (`'pyproject.toml'` does not exist) and python version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,10 +17,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
         with:
           path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml') }}
+          key: ${{ runner.os }}-{{ matrix.python-version }}-pip-${{ hashFiles('requirements.txt') }}-${{ hashFiles('dev-requirements.txt') }}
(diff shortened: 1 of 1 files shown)


hugovk:
We can actually remove `actions/cache` as a separate run and use the built-in pip caching form `actions/setup-python`:

https://github.com/actions/setup-python#caching-packages-dependencies

Something like this:

```yml
- uses: actions/setup-python@v3
  with:
    python-version: ${{ matrix.python-version }}
    cache: 'pip'
    cache-dependency-path: |
      requirements.txt
      dev-requirements.txt
```

---

This caching of course only applies to `actions/setup-python`, and not `deadsnakes/action` used for the `*-dev` versions. But:

1) we're not actually testing any `-dev` versions

2) if we do want to test `-dev`, `actions/setup-python` now also supplies them. One important difference: `actions/setup-python` supplies the latest alpha/beta/RC, whereas `deadsnakes/action` gives a nightly build. Usually the latest alpha/beta/RC is enough? If so we can remove:

```yml
      - uses: deadsnakes/action@v2.1.1
        if: endsWith(matrix.python-version, '-dev')
        with:
          python-version: ${{ matrix.python-version }}
```

---

Finally, we can update most (all?) of GitHub's own `actions/*` from `v2` to `v3`.

Mariatta (approved):

PR #409, opened 2022-02-23
Delete codecov.yml

sobolevn:
I want to discuss whether or not we need it.

First of all, looks like it is ignored by `codecov-action`: https://github.com/codecov/codecov-action/blob/b049ab51f46ef6cd9c7ca52a856631a818969c7c/src/buildExec.ts#L19-L156

Other users report it as well: https://github.com/codecov/codecov-action/issues/554

Secondly, it looks like our action works without it.
Run: https://github.com/sobolevn/bedevere/pull/2

So, I see two options:
1. Just delete it
2. Think of options we need to transfer to GitHub Actions `with:` arguments


Mariatta (approved):

Mariatta:
Thanks! Let's try it!

PR #410, opened 2022-02-23
Travis is not used anymore

sobolevn:


hugovk (approved):
Thanks, confirmed [GitHub Actions](https://github.com/python/bedevere/actions) is used for CI and there's no other mention of Travis in the code.

PR #399, opened 2022-01-27
Test against Python 3.11

Mariatta:

PR #416, opened 2022-03-20
Workaround to fix Python 3.11

hugovk:
Updates PR https://github.com/python/bedevere/pull/399 using workarounds from https://github.com/python/bedevere/pull/399#issuecomment-1073356616.


Mariatta (approved):
Thank you!

PR #419, opened 2022-03-24
Use the new "type" labels after GH migration.

Mariatta:
Closes https://github.com/python/core-workflow/issues/432

### DIFF
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -4,18 +4,19 @@ import pathlib
 
 from . import util
 
-LABEL_PREFIX = "type"
+TYPE_LABEL_PREFIX = "type"
 
 
 @enum.unique
 class Category(enum.Enum):
     """Category of Pull Request."""
-    bugfix = f"{LABEL_PREFIX}-bugfix"
-    documentation = f"{LABEL_PREFIX}-documentation"
-    enhancement = f"{LABEL_PREFIX}-enhancement"
-    performance = f"{LABEL_PREFIX}-performance"
-    security = f"{LABEL_PREFIX}-security"
-    tests = f"{LABEL_PREFIX}-tests"
+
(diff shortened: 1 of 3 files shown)


bedevere/prtype.py
```diff
@@ -4,18 +4,19 @@
...
-    performance = f"{LABEL_PREFIX}-performance"
-    security = f"{LABEL_PREFIX}-security"
-    tests = f"{LABEL_PREFIX}-tests"
+
+    type_bug = f"{TYPE_LABEL_PREFIX}-bug"
+    documentation = "docs"
+    type_feature = f"{TYPE_LABEL_PREFIX}-feature"
+    performance = "performance"
+    type_security = f"{TYPE_LABEL_PREFIX}-security"
+    tests = "tests"
```
ezio-melotti, line 19: `type-crash` is missing from the list, but it probably doesn't matter.

ezio-melotti (commented):
LGTM.

Do you think there's any downside applying it now or is it better/safer to wait?  I can still update the labels on my end so that they match the new labels during the migration.

On a more general note, many of the new labels are related to specific dirs/files in the repo, so bevedere will probably need some refactoring so that it can add more labels (also because `docs`, `performance`, `tests` are no longer types).

bedevere/prtype.py
Mariatta: Yeah I think bedevere really only cares about the `docs` and `tests` types at this moment. It doesn't do anything for the other type- labels.

Mariatta:
I would suggest changing the labels first on the CPython repo, and then deploy bedevere afterwards (by merging this PR).

ezio-melotti:
That makes sense, should we do it now?

Mariatta:
Yup let's do it!

ezio-melotti:
Thanks for the PR!
I renamed the labels on `python/cpython` and merged it.

PR #423, opened 2022-04-10
Support news files with a `gh-issue` substring

brettcannon:


Mariatta (approved):
Thanks I've added a commit for the test coverage.

Mariatta:
I'm also traveling today, my plane is about to take off. If anything breaks, I'll just have to fix it when I landed back in Canada.

PR #424, opened 2022-04-10
Fix gh-issue- prefix

JelleZijlstra:
Blurb generates gh-issue- but Bedevere only accepted gh-. This was discussed here: https://github.com/python/core-workflow/pull/431#discussion_r827293459


ambv:
We want it the other way: to fix Blurb instead. gh-issue- was meant to be present there only in the filename of a blurb news entry. Not anywhere else.

ambv:
It's easier for us to leave Blurb use gh-issue-*.

PR #420, opened 2022-03-25
Support GitHub Issues and drop b.p.o

Mariatta:
Expect gh- issue number in titles.
Remove logic for adding bpo url in the body.
Validate that the GH- issue is indeed an issue, not PR

Closes https://github.com/python/bedevere/issues/418

PR #431, opened 2022-04-13
Try to restore codecov

ezio-melotti:
It appears that codecov stopped working after #420.  Since that PR contained a change related to it, I'm reverting it to test if it resume working.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,0 +1,1 @@
+comment: true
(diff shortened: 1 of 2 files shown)


ezio-melotti:
The checks run with the default action, and they didn't trigger a codecov report.  I'm going to re-run them manually using the updated `.yml` to see if it works.

PR #436, opened 2022-04-13
Remove the follow-up about commit message using # instead of GH-

Mariatta:


methane:
Was there some context about this change?
Don't we need to replace `#` with `GH-` for now?

Issue #441, opened 2022-04-25
Update Codecov global YAML


hugovk:
Suggestion from #432:

---

Tom from Codecov here. Just saw this PR and wanted to offer my help in case you ran into any issues.

Looking at the global yaml, I'd recommend changing to
```yaml
comment: false
coverage:
  status:
    changes: false
    patch:
      default:
        informational: true
    project:
      default:
        informational: true
```
to get non-blocking status checks on PRs

_Originally posted by @thomasrockhu-codecov in https://github.com/python/bedevere/issues/432#issuecomment-1101448522_

---

The global YAML can be updated at https://app.codecov.io/account/gh/python/yaml

hugovk:
I've updated the YAML to the above. For reference, here is what it was before:

```yaml
comment: false
coverage:
  status:
    changes: false
    patch: false
    project: false
```

hugovk:
I created https://github.com/python/bedevere/pull/442 and coverage was sent to Codecov: https://codecov.io/gh/python/bedevere/commit/98bca146646c8bd29f3b6cd25c6dc4193e412046/

But no comment was created.

The repo`.github/.codecov.yml` is:

```yaml
comment: true
```

https://app.codecov.io/gh/python/bedevere/commit/98bca146646c8bd29f3b6cd25c6dc4193e412046 shows the combined YAML:

```yaml
codecov:
  require_ci_to_pass: true
comment: true
coverage:
  precision: 2
  range:
  - 70.0
  - 100.0
  round: down
  status:
    changes: false
    default_rules:
      flag_coverage_not_uploaded_behavior: include
    patch:
      default:
        informational: true
    project:
      default:
        informational: true
github_checks:
  annotations: true
```

@thomasrockhu-codecov Do you know what we're missing?

thomasrockhu-codecov:
@hugovk nothing comes up immediately to me. I'll need a little time to dig in to figure this out if you don't mind.

hugovk:
Thank you!

thomasrockhu-codecov:
@hugovk, sorry it took so long to get back, could you try removing `comment: true` from the yaml? This is a bug on our end with the way we are parsing the yaml.

hugovk:
Thanks, done in PR https://github.com/python/bedevere/pull/444. I deleted the `.github/.codecov.yml` file because it only contains `comment: true`.

There's no comment on that PR, does it need merging first?

thomasrockhu-codecov:
[...]

PR #442, opened 2022-04-25
CI: Cache pip via setup-python

hugovk:
The https://github.com/actions/setup-python action can now take care of caching pip for us, so we don't need an explicit https://github.com/actions/cache step.

Docs: https://github.com/actions/setup-python#caching-packages-dependencies

(This partly a test for https://github.com/python/bedevere/issues/441)


ezio-melotti:
This looks like a duplicate of #465.

PR #444, opened 2022-04-27
Delete .codecov.yml and "comment: true"

hugovk:
Delete the `comment: true` from the YAML to hopefully workaround a Codecov bug and enable commenting on this repo.

As suggested in https://github.com/python/bedevere/issues/441#issuecomment-1110357378

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,1 +1,0 @@
-comment: true


Mariatta:
I will mergie to see if it would solve the problem.

PR #446, opened 2022-04-29
Codecov: keep some 'comment' value to override global YAML

hugovk:
Fixes https://github.com/python/bedevere/issues/441.

Following suggestion at https://github.com/python/bedevere/issues/441#issuecomment-1112871092

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,2 @@
+comment:
+  layout: "reach, diff, flags, files"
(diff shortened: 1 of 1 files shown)


ezio-melotti (commented):
It would be useful to add a comment explaining why this is necessary, possibly pointing to the issue, since it's not entirely obvious.

hugovk:
Updated!

ezio-melotti (approved):

PR #461, opened 2022-05-30
Avoid running CI twice on PRs.

ezio-melotti:
I noticed that tests are run twice on each PR:
image

This change avoids that by only running them once whenever a PR is created/updated, and then once it is merged with master.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,12 @@
 name: Tests
 
-on: [push, pull_request, workflow_dispatch]
-
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+  
 env:
   FORCE_COLOR: 1
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: Tests
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
   workflow_dispatch:
   


.github/workflows/ci.yml
```diff
@@ -1,7 +1,12 @@
 name: Tests
 
-on: [push, pull_request, workflow_dispatch]
-
+on:
+  push:
+    branches:
+      - master
```
ezio-melotti, line 6:
```suggestion
      - main
```

nightlark:
> > If I understand correctly, you are saying that by limiting the `push` trigger to `main`, if you create a branch on your fork and push changes to your branch, it won't trigger the workflow because it's not a push on `master` and it not part of a PR yet, right?
> 
> Exactly.

While it is accurate to say that it won't trigger automatically on pushes,  it has the `workflow_dispatch` trigger, so you are still able to trigger tests on other branches when you want to test them in your fork.

ci-workflow-dispatch-trigger

hugovk:
Sure, that's also possible, but it's still an extra impediment. Making the human do more work so the computer can do less.

In your screen capture I see some 9 mouse clicks taking about 25 seconds! And then repeat for each push. The tests take around 25s, so could have finished by then.

And these tests are so quick, how much of a problem is it really for them to be run twice?

(For example, see also https://github.com/python/peps/pull/2215#issuecomment-1007914209)

nightlark:
It's not 9 clicks -- to be fair, I did accidentally click on a branch that I didn't intend to, which cuts it down to 7. If you anticipate running lots of tests, you can always keep the `Tests` page open and cut it down to ~3 clicks.

The [`gh` cli command](https://cli.github.com/manual/gh_workflow_run) also supports triggering workflow runs from the command line, so you could do it without leaving your shell window; it can also be used to view the result of workflow runs, so you don't need to switch to a browser or use any clicks to check the status of a workflow run in a browser either.

hugovk:
> It's not 9 clicks -- to be fair, I did accidentally click on a branch that I didn't intend to, which cuts it down to 7. If you anticipate running lots of tests, you can always keep the `Tests` page open and cut it down to ~3 clicks.

And that's an argument to keep them automated, humans can and will accidentally click the wrong thing :)

nightlark:
True. I'd just noticed this from an email notification on https://github.com/python/bedevere/issues/463#issuecomment-1141215468 and thought that the arguments against #461 sounded like they are all addressed by adding a workflow `workflow_dispatch` trigger -- then noticed in the changed files here that it already exists, so wanted to clear up that there exists a working alternative and point out where the feature is for anyone who isn't familiar with it.

Anyway, I think we are probably bikeshedding at this point given the time scales involved, small build matrix size, and the frequency of changes to this repository. As far as the change proposed in this PR goes from the perspective of a potential contributor to bedevere, I don't have a preference one way or the other for every commit triggering CI builds.

ezio-melotti:
For reference, we discussed this on Discord, and this is a summary of our findings and the options we considered:

* Currently the `on: ...` can be filtered [by branch, paths, or tags](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-excluding-branches), but not by repository or owner.
  * There is a [feature request on `github/feedback`](https://github.com/github/feedback/discussions/9098) that would allow to include/ignore forks in the `on: ...`
  * Using a special prefix in the branch name is an option -- e.g. [`pytest` uses the `test-me-` prefix](https://github.com/pytest-dev/pytest/blob/c533015c56f9d966543f2b660f4d10b802ec18ae/.github/workflows/test.yml#L8)
  * Using tags is also an option, in addition or in combination to branch prefixes
  * It's difficult to know/remember the naming conventions used by each repo
* It is possible to [use conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution), however this will still show the duplicate tests as skipped.
  * the `if: ...` can be used for [all jobs or individual steps](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)
  * `github.repository`, `github.repository_owner`, `github.event_name` can be used to filter
  *  e.g.:  `if: github.repository_owner != 'python' && github.event_name != 'push'`
* It is possible to use the ["skip duplicate actions"](https://github.com/marketplace/actions/skip-duplicate-actions) action.
* It is possible to change the workflow file in the fork, however
  * if it's changed on `main`, it makes it more difficult to keep it in sync with upstream
  * if it's changed in the branch, it then needs to be reverted before creating the PR
* It is possible to [use the `workflow_dispatch` as showed above](https://github.com/python/bedevere/pull/461#issuecomment-1141289876) [^1]

---

It seems that the options are:
1. close the PR and keep the duplicate checks
2. use `if: ...` to skip the duplicated jobs, even if they show up in the list of checks as skipped
3. add the `test-me-` branch prefix like `pytest` does (and possibly a `test-me` tag too)

Using the "skip duplicate actions" seems a bit overkill to me, and using the `workflow_dispatch` is a bit cumbersome (even though it could be configured as an additional option).

---


[^1]: Actually I think it's more complicated than that.  The video runs the tests on `main` using the workflow from a branch.  It is possible to add a field where to enter the branch that should be tested though, using an `input:` field under the `workflow_dispatch` and then `with: ref: ${{ github.event.inputs.selected_branch }}`.  Note that the branch name also has to be typed in, unless there is a "branch dropdown" input that I'm not aware of.

nightlark:
> ## Footnotes
> 1. Actually I think it's more complicated than that.  The video runs the tests on `main` using the workflow from a branch.  It is possible to add a field where to enter the branch that should be tested though, using an `input:` field under the `workflow_dispatch` and then `with: ref: ${{ github.event.inputs.selected_branch }}`.  Note that the branch name also has to be typed in, unless there is a "branch dropdown" input that I'm not aware of. [↩](#user-content-fnref-1-ab4b95adacbfa10f0357e79b1ba6fbfd)

It uses both the workflow and checks out the code from the selected branch -- here's the line from the job run in the screen capture showing the branch being checked out https://github.com/nightlark/bedevere/runs/6657800671?check_suite_focus=true#step:2:133

CAM-Gerlach:
FWIW, seems like either merging, moving or removing the upstream branch here is the only blocker to fully implementing python/core-workflow#460 on this last repo.

AlexWaygood:
FWIW I generally prefer the setup @ezio-melotti is proposing here for my own projects. But I agree with @hugovk that, given the relative infrequency of PRs to the bedevere repo, there isn't really a significant problem being solved here.

Given Hugo's strong objection, let's say that the status quo wins this one :)

arhadthedev:
@CAM-Gerlach Done.

PR #472, opened 2022-06-20
Build against pre-releases then stable releases.

jaraco:
I use this technique in jaraco/skeleton (and derived projects) to reduce the number of commits necessary to adopt a Python version by 50% (by adopting each version during pre-release and then switching to stable once final is released). In https://github.com/actions/setup-python/issues/213, I've reported the issue and I'm hoping that soon, "N.NN" will be sufficient to have the same behavior without having to require adding '-dev' and removing it later.

Feel free to reject if the ugliness outweighs the benefit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,13 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11-dev"]
+        python-version:
+        # Build on pre-releases until stable, then stable releases.
+        # actions/setup-python#213
+        - ~3.8.0-0
+        - ~3.9.0-0
+        - ~3.10.0-0
+        - ~3.11.0-0
 
     steps:
       - uses: actions/checkout@v3


.github/workflows/ci.yml
```diff
@@ -13,7 +13,13 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11-dev"]
+        python-version:
+        # Build on pre-releases until stable, then stable releases.
```
ezio-melotti, line 17: I think this comment could be clearer.
IIUC this syntax is a workaround that will pick the latest available version (which is what we want), and the version could be an alpha, a beta, or a final/stable release.

hugovk:
> I use this technique in jaraco/skeleton (and derived projects) to reduce the number of commits necessary to adopt a Python version by 50% (by adopting each version during pre-release and then switching to stable once final is released). In [actions/setup-python#213](https://github.com/actions/setup-python/issues/213), I've reported the issue and I'm hoping that soon, "N.NN" will be sufficient to have the same behavior without having to require adding '-dev' and removing it later.
> 
> Feel free to reject if the ugliness outweighs the benefit.

I think it's fine to add `3.11-dev` during pre-release, and then change to `3.11` on final release. That's only one extra commit and the config is much easier to understand.

The good news: there's an open PR to fix https://github.com/actions/setup-python/issues/213 that would allow us just to say `3.11` and it will test pre-release or stable, whichever is newest: https://github.com/actions/setup-python/pull/414

.github/workflows/ci.yml
jaraco: I copied this comment verbatim from jaraco/skeleton. I agree, the comment could provide a whole narrative about why it exists, but I try to be concise, to describe as briefly as possible the reason for the syntax, and then link to an issue that provides more context and tracks ultimate resolution.

If you'd like to suggest different wording, I'd be willing to apply that here and in jaraco/skeleton.

jaraco:
> I think it's fine to add `3.11-dev` during pre-release, and then change to `3.11` on final release. That's only one extra commit and the config is much easier to understand.

Is it fine? I know for myself, I maintain hundreds of different projects, so each change requires a new commit in each repo, and leaves that repo stale until I get around to making that commit. Honestly, I'd like to get to a place where I can make zero changes to the repo and it will automatically pick up new Python versions, but cutting in half the number of commits required is a real time saver for me.

> The good news: there's an open PR to fix [actions/setup-python#213](https://github.com/actions/setup-python/issues/213) that would allow us just to say `3.11` and it will test pre-release or stable, whichever is newest: [actions/setup-python#414](https://github.com/actions/setup-python/pull/414)

The bad news is that the maintainers are pushing back and insisting that the ugly workaround presented above is preferable. I hope by sharing this current-best-option and linking to the request, it will increase visibility and maybe help motivate a proper fix.

Of course, this project doesn't need to be a part of my advocacy, so feel free to reject.

I acknowledge, this syntax can and should go away if #214 is merged and released and applied here.

.github/workflows/ci.yml
```diff
@@ -13,7 +13,13 @@ jobs:
...
+        # actions/setup-python#213
```
ezio-melotti, line 18: Maybe something like this?  I also expanded the link to make it easier to open.
```suggestion
        # This syntax will select the latest available releases
        # including alpha, then beta, and finally stable releases.
        # See https://github.com/actions/setup-python/issues/213
```

ezio-melotti:
> I acknowledge, this syntax can and should go away if https://github.com/actions/setup-python/pull/414 is merged and released and applied here.

This is now merged and the workaround no-longer needed, so this PR can probably be closed.  Note that there is an unrelated and ongoing issue about 3.12 (see #542), but after that is resolved we should be able to just use `3.12` in the workflow.

PR #471, opened 2022-06-20
Add tox config and build using CI

jaraco:
Fixes #469.

- Add tox.ini from jaraco/skeleton
- Trim irrelevant details from tox.ini
- In CI, invoke tests with tox. Move pytest args to pytest config.

PR #474, opened 2022-06-20
Fix workflow syntax

hugovk:
Fixes https://github.com/python/bedevere/pull/471#issuecomment-1160538865.

Docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env


jaraco:
Thanks @hugovk for fixing my mistake. Looks like I got ahead of myself in the merge.

PR #475, opened 2022-06-20
Set env vars in tox to build 3.11 dependencies

hugovk:
After I pushed the tox changes (https://github.com/python/bedevere/pull/471, https://github.com/python/bedevere/pull/474) to my branch, it failed on Python 3.11:

* https://github.com/hugovk/bedevere/runs/6969338848?check_suite_focus=true

But it passed here:

* https://github.com/python/bedevere/actions/runs/2529801953

The problem is we are missing the env vars to temporarily fix the dependency build for 3.11. Let's move them out of the workflow and into tox.ini, so they're also available locally.

The reason it passed here was because there was still a pip cache from last time, whereas the cache had expired on my fork and it needed to build again from scratch.

Here's a passing run of this fix on my fork:

* https://github.com/hugovk/bedevere/actions/runs/2529924094

---

Also keep the colour output for pytest when run inside tox by passing in `FORCE_COLOR`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,14 +27,6 @@ jobs:
       - name: Install tox
         run: |
           python -m pip install tox
-        env:
-          # TEMP for 3.11
-          # https://github.com/aio-libs/aiohttp/issues/6600
-          AIOHTTP_NO_EXTENSIONS: 1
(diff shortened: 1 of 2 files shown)


ezio-melotti (approved):

jaraco:
I was noticing this issue too and was about to suggest the same change.

PR #478, opened 2022-06-22
Use the Python 3.10.5 runtime

Mariatta:


ezio-melotti (commented):
Can't this be automated in some way?
Will using `python-3.10` automatically pick the latest version?

Mariatta:
I don't think that syntax is supported (yet). I know I wished for a better way to do it.
This was brought up in https://github.com/heroku/heroku-buildpack-python/pull/899

Not sure if there's been any development about that from Heroku.

PR #483, opened 2022-06-30
Ignore case when matching backport PR title

Mariatta:


tests/test_backport.py
```diff
@@ -106,13 +106,14 @@ async def test_missing_backport_label():
     assert gh.delete_url is None
 
 
-async def test_backport_label_removal_success():
+@pytest.mark.parametrize('pr_prefix', ['GH', 'gh'])
```
ezio-melotti, line 109: Instead of repeating the `@pytest.mark.parametrize` you could create a parametrized fixture like:
```py
@pytest.fixture(params=['gh', 'GH'])
def pr_prefix(request):
    return request.param
```
and then just pass it directly as an arg to all the tests that need it.

Mariatta: Thanks, I refactored in https://github.com/python/bedevere/pull/484

PR #485, opened 2022-07-07
Auto add the ``skip news`` label to ``docs`` PRs

### LINKED ISSUE #457: Automatically apply "skip news" when detecting a docs-only change?
I noticed this in python/cpython#92877: bedevere seems to detect a docs-only change as it applies the "docs" label. But the bedevere/news step still fails as "skip news" is not applied. As docs changes don't need NEWS entries, maybe it would make sense to apply this label automatically and don't the the action step? But maybe there are cases that still require a NEWS entry and that would be missed with this change?

DanielNoord:
Closes #457.
Also closes https://github.com/python/core-workflow/issues/363.

Follow up can also tackle https://github.com/python/core-workflow/issues/421 but I decided to keep this small since the feedback on https://github.com/python/core-workflow/issues/363 shows that the consensus reached in #457 is not shared by everybody.

Two edge cases which were not considered in the original issue but on which somebody should probably take a decision:
1) What if the PR already has a news item? Currently it still adds the `skip news` label as shown by `test_docs_and_news`.
2) Should the `skip news` label also be applied if the PR has a test? Currently I thought I would not do so as to not make this change too controversial, but I'm happy to change this obviously.

Let me know what you think!

### DIFF
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -8,8 +8,8 @@ TYPE_LABEL_PREFIX = "type"
 
 
 @enum.unique
-class Category(enum.Enum):
-    """Category of Pull Request."""
+class Labels(enum.Enum):
+    """Labels that can be applied to a Pull Request."""
 
     type_bug = f"{TYPE_LABEL_PREFIX}-bug"
     documentation = "docs"
@@ -19,7 +19,7 @@ class Category(enum.Enum):
     tests = "tests"
 
 
-async def add_category(gh, issue, category):
+async def add_label(gh, issue, category):
     """Apply this type label if there aren't any type labels on the PR."""
     if any(label.startswith("type") for label in util.labels(issue)):
         return
@@ -47,7 +47,7 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         else:
             return
     if tests:
-        await add_category(gh, issue, Category.tests)
+        await add_label(gh, issue, Labels.tests)
     elif docs:
-        await add_category(gh, issue, Category.documentation)
+        await add_label(gh, issue, Labels.documentation)
     return
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -20,12 +20,12 @@ class Labels(enum.Enum):
     skip_news = "skip news"
 
 
-async def add_label(gh, issue, labels):
-    """Apply this type label if there aren't any type labels on the PR."""
-    if any(label.startswith("type") for label in util.labels(issue)):
-        return
-    label_names = [c.value for c in labels]
-    await gh.post(issue["labels_url"], data=label_names)
+async def add_labels(gh, issue, labels):
+    """Add the specified labels to the PR."""
+    current_labels = util.labels(issue)
+    label_names = [c.value for c in labels if c.value not in current_labels]
+    if label_names:
+        await gh.post(issue["labels_url"], data=label_names)
 
 
 async def classify_by_filepaths(gh, pull_request, filenames):
@@ -49,10 +49,10 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         else:
             return
     if tests:
-        await add_label(gh, issue, [Labels.tests])
+        await add_labels(gh, issue, [Labels.tests])
     elif docs:
         if news:
-            await add_label(gh, issue, [Labels.docs])
+            await add_labels(gh, issue, [Labels.docs])
         else:
-            await add_label(gh, issue, [Labels.docs, Labels.skip_news])
+            await add_labels(gh, issue, [Labels.docs, Labels.skip_news])
     return
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -12,7 +12,7 @@ class Labels(enum.Enum):
     """Labels that can be applied to a Pull Request."""
 
     type_bug = f"{TYPE_LABEL_PREFIX}-bug"
-    documentation = "docs"
+    docs = "docs"
     type_feature = f"{TYPE_LABEL_PREFIX}-feature"
     performance = "performance"
     type_security = f"{TYPE_LABEL_PREFIX}-security"
@@ -49,5 +49,5 @@ async def classify_by_filepaths(gh, pull_request, filenames):
     if tests:
         await add_label(gh, issue, Labels.tests)
     elif docs:
-        await add_label(gh, issue, Labels.documentation)
+        await add_label(gh, issue, Labels.docs)
     return
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -19,11 +19,11 @@ class Labels(enum.Enum):
     tests = "tests"
 
 
-async def add_label(gh, issue, category):
+async def add_label(gh, issue, label):
     """Apply this type label if there aren't any type labels on the PR."""
     if any(label.startswith("type") for label in util.labels(issue)):
         return
-    await gh.post(issue["labels_url"], data=[category.value])
+    await gh.post(issue["labels_url"], data=[label.value])
 
 
 async def classify_by_filepaths(gh, pull_request, filenames):
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -17,13 +17,15 @@ class Labels(enum.Enum):
     performance = "performance"
     type_security = f"{TYPE_LABEL_PREFIX}-security"
     tests = "tests"
+    skip_news = "skip news"
 
 
-async def add_label(gh, issue, label):
+async def add_label(gh, issue, labels):
     """Apply this type label if there aren't any type labels on the PR."""
     if any(label.startswith("type") for label in util.labels(issue)):
         return
-    await gh.post(issue["labels_url"], data=[label.value])
+    label_names = [c.value for c in labels]
+    await gh.post(issue["labels_url"], data=label_names)
 
 
 async def classify_by_filepaths(gh, pull_request, filenames):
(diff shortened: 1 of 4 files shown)


DanielNoord:
Well, that's stupid of me. That `if ... elif ... if` can never be true. Sorry about the churn!

Questions from my OP are still relevant though!

DanielNoord:
Thanks for the speedy review 😄 

> 1. if a PR has a news item, the label shouldn't be applied (see inline comment)

Changed!

> 2. if a PR has only tests, I also think it shouldn't be applied (at least not in this PR)

This is the status quo, so no further changes necessary I think.

I merged https://github.com/python/bedevere/pull/486 into this as I think that should probably be merged before this. I can rebase after that has been merged.

DanielNoord:
Already rebased on #486 as I mixed up my branches, sorry about that.

Consolidated all changes + review of this PR in a single commit.

bedevere/prtype.py
```diff
@@ -17,13 +17,15 @@ class Labels(enum.Enum):
...
     type_security = f"{TYPE_LABEL_PREFIX}-security"
     tests = "tests"
+    skip_news = "skip news"
 
 
-async def add_label(gh, issue, label):
+async def add_label(gh, issue, labels):
     """Apply this type label if there aren't any type labels on the PR."""
     if any(label.startswith("type") for label in util.labels(issue)):
         return
```
ezio-melotti, line 26:
```suggestion
    """Add the specified labels to the PR."""
```
I think this part could be removed, since it's based on the assumption that all the labels passed are `type-*` labels and the fact that the PR can only have one `type-*` label, even though now it's being used for other labels too and `docs`/`tests`/`performances` are no longer `type-*` labels.

bedevere/prtype.py
```diff
@@ -17,13 +17,15 @@ class Labels(enum.Enum):
...
-    await gh.post(issue["labels_url"], data=[label.value])
+    label_names = [c.value for c in labels]
+    await gh.post(issue["labels_url"], data=label_names)
```
ezio-melotti, line 28:
```suggestion
    current_labels = util.labels(issue)
    label_names = [c.value for c in labels if c.value not in current_labels]
    if label_names:
        await gh.post(issue["labels_url"], data=label_names)
```

bedevere/prtype.py
```diff
@@ -17,13 +17,15 @@ class Labels(enum.Enum):
     performance = "performance"
     type_security = f"{TYPE_LABEL_PREFIX}-security"
     tests = "tests"
+    skip_news = "skip news"
 
 
-async def add_label(gh, issue, label):
+async def add_label(gh, issue, labels):
```
ezio-melotti, line 23:
```suggestion
async def add_labels(gh, issue, labels):
```

bedevere/prtype.py
```diff
@@ -47,7 +49,10 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         else:
             return
     if tests:
-        await add_label(gh, issue, Labels.tests)
+        await add_label(gh, issue, [Labels.tests])
```
ezio-melotti, line 52:
```suggestion
        await add_labels(gh, issue, [Labels.tests])
```

bedevere/prtype.py
```diff
@@ -47,7 +49,10 @@ async def classify_by_filepaths(gh, pull_request, filenames):
...
     elif docs:
-        await add_label(gh, issue, Labels.docs)
+        if news:
+            await add_label(gh, issue, [Labels.docs])
+        else:
+            await add_label(gh, issue, [Labels.docs, Labels.skip_news])
```
ezio-melotti, line 57:
```suggestion
        if news:
            await add_labels(gh, issue, [Labels.docs])
        else:
            await add_labels(gh, issue, [Labels.docs, Labels.skip_news])
```

ezio-melotti (changes requested):

bedevere/prtype.py
DanielNoord: I do not completely follow you. Isn't this a check that makes sure we are not overwriting the current label types if there are already specific `type-*` labels?

DanielNoord: Oh never mind, didn't see you other suggestion.

DanielNoord:
@ezio-melotti Your suggestions break one test. If have pushed a commit that fixes it, but could you look into whether this is correct?

https://github.com/python/bedevere/pull/485/commits/a0f850bcc8235ca55554109b8b34cd137803d30c

ezio-melotti:
I fixed the merge conflicts and left another review, however I'm still not too convinced about the `add_labels` function.

`prtype.classify_by_filepaths` is called whenever a new PR is opened, and as args it receives `gh` (a `GitHubAPI` object), the `pull_requests` and the `filenames`:
https://github.com/python/bedevere/blob/bf7a01b609f417bac4069e9b4eb9ecdf7487d631/bedevere/filepaths.py#L12-L21

<details>
<summary>(click to see the definition of `gh`)</summary>

https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/__main__.py#L35-L37
</details>

It then claim to categorize a PR:
https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/prtype.py#L29-L30

but first it retrieves the issue linked to the PR:
https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/prtype.py#L37

<details>
<summary>(or at least that's what it seems to do -- click to see source)</summary>

https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/util.py#L90-L92
</details>

then it passes the `issue` to `add_label(s)`:
https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/prtype.py#L49-L52

and `add_label(s)` claim to update the PR while working on the issue (it first checks the issue labels and then seemingly adds the labels to the issue):
https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/prtype.py#L22-L26

So it seems to me that the intention is to update the PR, but -- unless something is misnamed and `issue` is actually the `PR` -- the issue is being updated instead. 

Can someone more familiar with bedevere confirm?

DanielNoord:
> but first it retrieves the issue linked to the PR:
> 
> https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/prtype.py#L37
> 
> (or at least that's what it seems to do -- click to see source)
> https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/util.py#L90-L92
> 


I think the confusion comes from this function (which is poorly named imo). It's based on the following payload for `pull_request` events:
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-example-33

As you can see `issue_url` is just the `pull_request` URL in its "issue form". Basically for this PR that link would be:
https://api.github.com/repos/python/bedevere/issues/485

Rather `issue_for_pr` it should probably be called `api_entry_for_pr` or something like that.

DanielNoord:
> Maybe just something like `get_pr_info` or `get_pr_data` with a clearer docstring.

I looked into a refactoring PR, but it seems like this terminology is being used throughout the code base. Take for example these lines:
https://github.com/python/bedevere/blob/83a2d2273d3825297e2a09ea4024b84ca210439d/bedevere/backport.py#L61-L67

Again `issue` here seems to refer to the `issue_url`. Since this is in accordance with the Github API I think it might make sense to keep this as is and update the docstring of `issue_for_PR`. If you still want me to rename it let me know and I'll open another PR.


> I think the whole test should be rewritten.

Rewritten and added a new/old test to bump coverage again. We now check that 1) we don't post if there is an exact match with current labels, 2) we "partially" post if there is a partial match with the current labels.

bedevere/util.py
```diff
@@ -106,7 +110,11 @@ async def patch_body(gh, pull_request, issue_number):
     if not re.search(rf"(^|\b)(GH-|gh-|#){issue_number}\b", pull_request["body"]):
         return await gh.patch(
             pull_request["url"],
-            data={"body": BODY.format(body=pull_request["body"], issue_number=issue_number)},
+            data={
+                "body": BODY.format(
+                    body=pull_request["body"], issue_number=issue_number
+                )
+            },
```
ezio-melotti, line 117: This change seems unrelated.

ezio-melotti (changes requested):
The new tests look good.  I see what you are saying about the refactoring too.

I left a comment about the docstring, but I agree we shouldn't rename the function in this PR.

bedevere/util.py
```diff
@@ -88,7 +88,11 @@ async def files_for_PR(gh, pull_request):
 
 
 async def issue_for_PR(gh, pull_request):
-    """Get the issue data for a pull request."""
+    """Get the issue_url for a pull request.
+
+    The issue_url is the API endpoint for the given pull_request.
+    This terminology follows the official Github API and its documentation.
+    """
```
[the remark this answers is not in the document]
DanielNoord: Ah, my bad. You're right. What about:

```suggestion
    """Get the issue_url for a pull request.

    The issue_url is a JSON payload with all data relevant for the given pull_request.
    This terminology follows the official Github API and its documentation.
    """
```

bedevere/util.py
DanielNoord, line 117:
```suggestion
            data={"body": BODY.format(body=pull_request["body"], issue_number=issue_number)},
```

bedevere/util.py
```diff
@@ -88,7 +88,11 @@ async def files_for_PR(gh, pull_request):
...
     return await gh.getitem(pull_request["issue_url"])
```
ezio-melotti, line 96: Maybe this is enough (especially if we are going to refactor it later:
```suggestion
    """Return a dict with data about the given PR."""
    # "issue_url" is the API endpoint for the given pull_request (despite the name)
    return await gh.getitem(pull_request["issue_url"])
```

ezio-melotti (approved):

PR #487, opened 2022-07-08
Do not fail issue number test for all PRs with ``skip issue`` label

### LINKED ISSUE #468: Fails issue number check even with skip-issue set
I created a documentation improvement [PR with no issue number](https://github.com/python/cpython/pull/93644) and with `skip-issue` set. There was no problem with bedevere on this, and I successfully merged the PR in due course. I then set back-port flags for 3.10 and 3.11; back-port PRs were duly created and correctly inherited the `skip-issue` label. There were no problems with bedevere on these, either. The 3.11 back-port PR tests ran successfully and I merged it, too. The 3.10 back-port PR is where the problem occurred, though not initially. The tests for this failed due to a `suspicious.py` check and, to see if this was some transient issue, I closed the PR and opened it again just to re-run the tests. At this point, bedevere failed the PR on the issue number. The problem is that the back-port PR contains the text `(GH-96344)` from the original PR against `main`; for some reason, bedevere interprets this as an invalid issue number, when it would seem that it should check for the presence of `skip-issue` first. [This is the 3.10 back-port PR which shows the issue](https://github.com/python/cpython/pull/93648).

DanielNoord:
Closes #468.

This check should be done before we get any other information.

I had to change some test data because they were "incomplete". I could also check for the label on the old lines 52-54, but then we would unnecessarily validate the issue number we found in the title. Thought it made more sense to do it like this.

### DIFF
--- a/bedevere/gh_issue.py
+++ b/bedevere/gh_issue.py
@@ -33,11 +33,14 @@ async def set_status(
     event, gh: GitHubAPI, *args, session: ClientSession, **kwargs
 ):
     """Set the issue number status on the pull request."""
+    issue = await util.issue_for_PR(gh, event.data["pull_request"])
+    if util.skip("issue", issue):
+        await util.post_status(gh, event, SKIP_ISSUE_STATUS)
+        return
+
     issue_number_found = ISSUE_RE.search(event.data["pull_request"]["title"])
     if not issue_number_found:
-        issue = await util.issue_for_PR(gh, event.data["pull_request"])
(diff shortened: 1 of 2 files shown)


vsajip:
The change in `gh_issue.py` looks good to me - but I would need more time to get my head around the test code :slightly_frowning_face:

ezio-melotti (approved):

ezio-melotti:
This seems to work: https://github.com/python/cpython/pull/98163

Issue #440, opened 2022-04-24
Conflict while backporting issue-less PRs


ezio-melotti:
I was trying to backport python/cpython#32342 with `cherry_picker`, and:
* For 3.9/3.10 it added a message about the backport to the original PR, but complained in the checks that the issue number wasn't valid, so I had to remove it from the title
  * python/cpython#91890
  * python/cpython#91891
* For 3.7/3.8 it didn't add a message to the original PR, but also didn't complain even with the PR number in the title
  * python/cpython#91892
  * python/cpython#91893

All backport PRs had the `skip issue` (and `skip news`) labels.  I also noticed that the original PR didn't have the `needs backport to 3.7/3.8` labels, and after adding them and editing the title to retrigger the check, it added the message and failed the check.

I would expect it to:
* not fail the check if the PR has a `skip issue` label
* add a message to the original PR even if the `needs backport to *` label is missing

DanielNoord:
The first issue is a duplicate of #468 for which I have a fix in #487.

Should probably refocus this issue on the second issue about not posting a message on the original issue for some python versions.

PR #489, opened 2022-07-12
Comment about a backport even if the original PR doesn't have a label

DanielNoord:
Closes https://github.com/python/bedevere/issues/440.

The `if ... return` makes it so we never comment if there original PR doesn't have a back port label.


DanielNoord:
@ezio-melotti Gentle ping as you have been helpful in reviewing some of my other PRs and this is fixing an issue that you originally opened yourself. Feel free to ignore of course!

ezio-melotti (approved):

Issue #269, opened 2020-09-15
Move the PR back to awaiting review if review was dismissed


Mariatta:
In https://github.com/python/cpython/pull/22253#pullrequestreview-488290678, I approved the PR but later decided to retract my review.

I think in this situation the PR should be moved back to "awaiting review" state.

DanielNoord:
Implementation in https://github.com/python/bedevere/pull/490 😄

DanielNoord:
@Mariatta Gentle ping for a review of #490. Not sure if this is appreciated but since this affects `awaiting merge` labels I can imagine this could be more problematic than some of (my) other open PRs.

PR #490, opened 2022-07-12
Update stage based on dismissed reviews

DanielNoord:
Closes https://github.com/python/bedevere/issues/269

Also removed two unused imports.

### DIFF
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -178,6 +178,17 @@ async def core_dev_reviewers(gh, pull_request_url):
             yield reviewer
 
 
+async def reviewers(gh, pull_request_url):
+    """Find any type of reviewers."""
+    # Unfortunately the reviews URL is not contained in a pull request's data.
+    async for review in gh.getiter(pull_request_url + "/reviews"):
+        reviewer = util.user_login(review)
+        # Ignoring "comment" reviews.
+        actual_review = review["state"].lower() in {"approved", "changes_requested"}
+        if actual_review:
+            yield reviewer
+
+
 @router.register("pull_request_review", action="submitted")
 async def new_review(event, gh, *args, **kwargs):
     """Update the stage based on the latest review."""
@@ -223,6 +234,27 @@ async def new_review(event, gh, *args, **kwargs):
             await gh.post(pull_request["comments_url"], data={"body": comment})
         else:  # pragma: no cover
             raise ValueError(f"unexpected review state: {state!r}")
+
+
+@router.register("pull_request_review", action="dismissed")
+async def dismissed_review(event, gh, *args, **kwargs):
+    """Update the stage based on a dismissed review."""
+    pull_request = event.data["pull_request"]
+
+    # Poor-man's asynchronous any().
+    async for _ in core_dev_reviewers(gh, pull_request["url"]):
+        # No need to update the label as there is still a core dev review.
+        return
+    else:
+        async for _ in reviewers(gh, pull_request["url"]):
+            # Request review from core dev
+            await stage(
+                gh, await util.issue_for_PR(gh, pull_request), Blocker.core_review
+            )
+            return
+        else:
+            # Waiting for anybody to leave a review.
+            await stage(gh, await util.issue_for_PR(gh, pull_request), Blocker.review)
 
 
 @router.register("issue_comment", action="created")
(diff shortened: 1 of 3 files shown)


bedevere/stage.py
```diff
@@ -227,6 +236,27 @@ async def new_review(event, gh, *args, **kwargs):
...
+@router.register("pull_request_review", action="dismissed")
+async def dismissed_review(event, gh, *args, **kwargs):
+    """Update the stage based on a dismissed review."""
+    pull_request = event.data["pull_request"]
+
+    # Poor-man's asynchronous any().
+    async for _ in core_dev_reviewers(gh, pull_request["url"]):
+        # No need to update the label as there is still a core dev review.
+        return
+    else:
```
DanielNoord, line 248: This is else is necessary for coverage on `3.9`. On `3.10` and `3.8` this works without, but on `3.9` `coverage` complains about a missing branch.

ezio-melotti:
The chart in the README should also be updated.

DanielNoord:
@ezio-melotti How should this be represented in the chart. There is no single flow from the dismissed review as it is based on the highest "priority" remaining state. Can such a flow be represented in mermaid?

ezio-melotti:
Good question.  What are the new flows introduced by this PR?

If there are too many, it might be easier to add a note, either in the text after the graph, or within the graph.  One way to do that is `G[When a review is dismissed\nthe previous state is restored]`, even though it's probably abusing nodes.

Another option is to add something like `* -- review dismissed --> B` even though that would be inaccurate if dismissing a review might send to different nodes (e.g. `awaiting review` or `awaiting core review`).  In that case we would have to add one for each possible target, which might make the graph too crowded.

There are [workarounds to make arrow come from an hidden node](https://stackoverflow.com/a/73420433) so that only the label and the arrow are visible making the graph a bit less crowded.

This rendering shows all 3 options (arrow/label from `[ * ]` node, arrow/label from hidden node, note box).  I think the simplest solution is just the note box [^1].

```mermaid
flowchart TD
    A([New PR]):::creator
    A -- by contributor --> B[Awaiting review]:::anyone
    A -- by core dev --> C[Awaiting core review]:::coredev
    B & C -- new review by\nanother contributor --> C
    C & B & E  -- new core review\nrequests changes --> D[Awaiting changes]:::creator
    D -- changes by contributor --> E[Awaiting change review]:::coredev
    C & E & B -- new core review\napproves ---> F[Awaiting merge]:::coredev
    * -- review dismissed --> B
    H:::hidden -- review dismissed --> B
    G[When a review is dismissed\nthe previous state is restored]
classDef creator stroke:#CC0;
classDef anyone stroke:#00C;
classDef coredev stroke:#0C0;
classDef triager stroke:#C0C;
classDef hidden opacity:0%;
linkStyle 0,1,7 stroke:#CC0,color:auto;
linkStyle 2,3 stroke:#00C,color:auto;
linkStyle 4,5,6,8,9,10 stroke:#0C0,color:auto;
```

[^1]: which could also be expanded to include info about the colors :thinking:

DanielNoord:
> Good question. What are the new flows introduced by this PR?

It reverts to the "highest order" remaining state. So, if a core reviewer dismisses but a normal user did review we go to `awaiting core review`. If there is not review we go to `awaiting review` etc.


> This rendering shows all 3 options (arrow/label from `[ * ]` node, arrow/label from hidden node, note box). I think the simplest solution is just the note box [1](#user-content-fn-1-42cd1eef5a610a9100e86417fc6858ab).

Agreed, I think this makes the most sense. I'll add a note!

ezio-melotti:
The wording sounds good to me.  Do you know how many new requests this change introduces?
I'm not sure if that's a concern, but IIRC we might have a limited numbers of requests and by increasing them we might hit the limit.  If they are triggered only when a review is dismissed though, then I don't think it's going to be a problem since that happens rarely.

bedevere/stage.py
```diff
@@ -197,6 +206,27 @@ async def new_review(event, gh, *args, **kwargs):
             raise ValueError(f"unexpected review state: {state!r}")
 
 
+@router.register("pull_request_review", action="dismissed")
```
DanielNoord, line 209: @ezio-melotti In response to your comment, this is the only new registered route.

However, iterating on reviewers and core_reviewers might also be requests? I'm not sure which requests are counted towards the limit.

DanielNoord:
Is there anything I can do to help this PR move along? 😄

ezio-melotti (approved):

PR #492, opened 2022-07-14
Label a PR before posting the blurb help comment

DanielNoord:
Follow up to https://github.com/python/bedevere/pull/485

https://github.com/python/cpython/pull/94828 shows the previous PR works: changes to documentation are correctly labeled as `skip news`.
However, we're doing the check for the `blurb_it` help comment before we do the labeling.

This makes it so that we don't check whether we need to post that comment if we are going to label a PR as skip news.

/CC @ezio-melotti

### DIFF
--- a/bedevere/filepaths.py
+++ b/bedevere/filepaths.py
@@ -16,7 +16,9 @@ async def check_file_paths(event, gh, *args, **kwargs):
     pull_request = event.data['pull_request']
     files = await util.files_for_PR(gh, pull_request)
     filenames = [file['file_name'] for file in files]
-    await news.check_news(gh, pull_request, files)
     if event.data['action'] == 'opened':
-        await prtype.classify_by_filepaths(gh, pull_request, filenames)
-
+        labels = await prtype.classify_by_filepaths(gh, pull_request, filenames)
+        if prtype.Labels.skip_news not in labels:
+            await news.check_news(gh, pull_request, files)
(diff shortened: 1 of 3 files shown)


bedevere/filepaths.py
```diff
@@ -16,7 +16,9 @@ async def check_file_paths(event, gh, *args, **kwargs):
...
     filenames = [file['file_name'] for file in files]
-    await news.check_news(gh, pull_request, files)
     if event.data['action'] == 'opened':
-        await prtype.classify_by_filepaths(gh, pull_request, filenames)
-
+        labels = await prtype.classify_by_filepaths(gh, pull_request, filenames)
+        if prtype.Labels.skip_news not in labels:
+            await news.check_news(gh, pull_request, files)
+    else:
+        await news.check_news(gh, pull_request, files)
```
ezio-melotti, line 24: It seems to me that `news.check_news` is already checking for the `skip-news` labels, so simply swapping the order of the checks might be enough.  However, if we do that we might run into a race condition if `check_news` ends up checking the labels before we are done assigning them.  

Are you able to verify if this is an issue?  If not, we can just play it safe like you are doing now.

DanielNoord: According to the tests we need to take the safe route. Could this be because the asynchronous call to the `check_news` is too fast to wait for the labels to be updated?

PR #507, opened 2022-09-28
Use actual precise Python version @ coverage flags

webknjaz:
The matrix factor for the Python version is imprecise and doesn't always contain characters suitable for Codecov flags. `actions/setup-python@v4` now provides an output specifying exactly what's been installed based on the request. This value is perfect for use in flags. This patch implements just that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        id: python-install
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
@@ -37,4 +39,4 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
-          flags: Python_${{ matrix.python-version }}
+          flags: Python_${{ steps.python-install.outputs.python-version }}


ezio-melotti:
This change creates duplicated entry in the table included in the previous message, with the `matrix` versions showing `?` and the actual versions showing the right coverage.  Is this something that can be fixed?

On [codecov](https://app.codecov.io/gh/python/bedevere/pull/507) only the actual versions are included.

webknjaz:
@ezio-melotti

> On [codecov](https://app.codecov.io/gh/python/bedevere/pull/507) only the actual versions are included.

Looks like that comment includes historic flags that existed before this PR, showing that this PR's workflow jobs didn't report anything with those flags. I'm going to guess that it compares what changed vs `main` and that's why they are in the table. I suppose they should disappear once there's a commit on master that no longer reports those flags.

hugovk:
So (for example) `Python_3.9` is the old one and `Python_3.9.14` is the new one.

Do we actually care about the patch version in the coverage history?

That is, when checking the coverage, do we want continuous history across all 3.9 versions or does it not matter?

webknjaz:
> That is, when checking the coverage, do we want continuous history across all 3.9 versions or does it not matter?

I think this may be an important bit of information. Especially, if there's some conditional behavior/hacks that are only triggered on some patch versions. This might prove useful in future archeology while debugging stuff.

hugovk:
Fair enough! I suppose we could even send x.y and x.y.z flags, but will leave it up to you.

hugovk (approved):

webknjaz:
There's no easy way to infer that x.y version without resorting to an extra task. This is because the matrix values aren't always suitable for flags.

ezio-melotti (approved):

Issue #501, opened 2022-08-18
Add PR links to the issue


ezio-melotti:
Currently, `bedevere` adds a convenient link to issue `gh-NNNNN` to the first message of a PR that includes `gh-NNNNN` in the title when the PR is created:
![image](https://user-images.githubusercontent.com/25624924/185292198-8b42cb57-0c75-4d77-9ab6-97f44de032a8.png)

However, if you look at the [corresponding issue](https://github.com/python/cpython/issues/95853), the link to the PR is lost in the issue timeline among comments, metadata changes, and mentions from other issues and PRs:
<details>
<summary>Click to see an excerpt of the issue timeline</summary>

![image](https://user-images.githubusercontent.com/25624924/185292864-2ac405b1-10e0-4c75-8c45-a20f51eced53.png)
</details>

The only way to find related PRs is to scroll through the whole issue history looking for mentions, and figure out which of these are PRs that solve the issue.

---

When a PR that includes `gh-NNNNN` in the title is created `bedevere` should:
* add a link to the issue in the first message of the PR (this already works)
* add a link to the PR in the first message of the issue

If more PRs that include the same `gh-NNNNN` are created (e.g. backports), the list of PRs in the first message of the issue should be updated accordingly.  This will provide an easy way to track all related PRs.

---

About the implementation:
* only PRs that include the issue in the title should be added to the list
* PRs/issues that simply mention this issue in comments could be added to separate lists, but that's out of the scope of this issue
* the PRs could be listed using a checklist (or even better, the upcoming improved tasklists)
* a short notice (possibly with a link) could be added to clarify that the the list wasn't added by the OP, but by `bedevere`
* this could replace the messages that `bedevere` adds when backport PRs are created

zware:
GitHub does now offer actual links between issues and PRs (see the "Development" section in the right sidebar of this or other issues).  However, I have yet to figure out how it's actually supposed to work; the dropdown menu frequently fails to offer me anything approaching relevant, and the magic linking words to include in a message body really are quite magical.  If we could that linking working reliably (possibly with help from Sir Bedevere), we could hopefully move away from having to have Bedevere edit any messages.  There may be some useful documentation on this somewhere, but I have failed to find it after not looking at all :)

ezio-melotti:
I created this issue exactly because that feature seems to be kind of useless for our use case.  As far as I understand:
[...]

PR #511, opened 2022-10-26
Add Pull Request Link to Issue

saadmk11:
closes https://github.com/python/bedevere/issues/501

### DIFF
--- a/bedevere/gh_issue.py
+++ b/bedevere/gh_issue.py
@@ -24,6 +24,10 @@ ISSUE_URL: Dict[IssueKind, str] = {
     "gh": "https://github.com/python/cpython/issues/{issue_number}",
     "bpo": "https://bugs.python.org/issue?@action=redirect&bpo={issue_number}"
 }
+ISSUE_CHECK_URL: Dict[IssueKind, str] = {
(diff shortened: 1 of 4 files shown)


ezio-melotti (commented):
Does this handle issues with multiple PRs?

saadmk11:
> Does this handle issues with multiple PRs?

If you mean multiple PR's pointing to the same issue, then the answer is yes. :)

saadmk11:
> I really like the way this feature is working! Thanks @saadmk11!
> 
> See [python/cpython#98852](https://github.com/python/cpython/issues/98852) for a nice example of it in action

Always happy to contribute. This was my first contribution to `bedevere` repository. Would be happy to contribute to some more interesting features. :)

ezio-melotti:
> I really like the way this feature is working! Thanks @saadmk11!

Indeed, thanks for the PR!

There are more possible enhancement that could be added in future PRs.  One would be adding the issue to the same list, instead of adding a new one-element list for each PR.  Another option is to add them to a checklist or (even better) to a tasklist.
@saadmk11, would you be interested in expanding this feature?  If not, I can open a separate issue as a reminder.

Issue #517, opened 2022-11-06
Improve formatting of PR links added to issues


ezio-melotti:
This is a follow-up of:
* #501 
* #511

Currently the code generated when multiple PRs are linked to an issue looks like:
```html
<!-- gh-pr-number: gh-98920 -->
* PR: gh-98920
<!-- /gh-pr-number -->


<!-- gh-pr-number: gh-98952 -->
* PR: gh-98952
<!-- /gh-pr-number -->


<!-- gh-pr-number: gh-98969 -->
* PR: gh-98969
<!-- /gh-pr-number -->
```

And the output looks like:
![image](https://user-images.githubusercontent.com/25624924/200157295-5e8a03fc-94e8-4a6e-90d9-b189d7ad71c2.png)

It would be better to:
1. combine these into a single multi-item list, rather than multiple single-item lists
2. use a checklist or tasklist instead of a regular bullet list

If the PRs are combined in a single list, we won't be able to just append a new PR to the end, but we will need to add the PR link before the closing tag (`<!-- /gh-pr-number -->`).  In addition, the opening tag lists the PR number, so that will need to be updated too -- unless we remove the number from the tag, since it's already included in the list.

Using tasklists is probably better than checklists, and for that, the generated source should look like:
`````md
<!-- gh-linked-prs -->
``` [tasklist]
### Linked PRs
- [ ] gh-98920
- [ ] gh-98952
- [ ] gh-98969
```
<!-- /gh-linked-prs -->
`````
(there should be no space between ` ``` ` and `[tasklist]`, I added it here otherwise the output breaks).

Since tasklists support a title (e.g. `### Linked PRs`), there is no need to repeat `PR:` before each PR number.  (Note that tasklists are a beta feature enabled for the `python` org, and don't work for other orgs (yet).)

(cc @saadmk11)

saadmk11:
Hi @ezio-melotti, Thanks for creating such a detailed issue. I have added a PR with the required changes here:

- https://github.com/python/bedevere/pull/518

As I do not have access to any **repo/org** with the **tasklist beta** feature enabled, I was only able to test the generated markdown but I was not able to test how it looks after rendering. (This PR generates markdown exactly same as the markdown you provided in the Issue body)

PR #518, opened 2022-11-07
Add PR Links as Task List Items to Issues

saadmk11:
closes https://github.com/python/bedevere/issues/517

### DIFF
--- a/bedevere/util.py
+++ b/bedevere/util.py
@@ -38,7 +38,7 @@ ISSUE_BODY_TASK_LIST_TEMPLATE = f"""\n
 ISSUE_BODY_TASK_LIST_PATTERN = re.compile(
     rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\r?\n```\[tasklist\])"
     rf"(?P<tasks>.*?)"
-    rf"(?P<end>```\s{{1,2}}{ISSUE_BODY_CLOSING_TAG})",
+    rf"(?P<end>```\r?\n{ISSUE_BODY_CLOSING_TAG})",
     flags=re.DOTALL
 )
 
--- a/bedevere/util.py
+++ b/bedevere/util.py
@@ -36,8 +36,9 @@ ISSUE_BODY_TASK_LIST_TEMPLATE = f"""\n
 
 # Regex pattern to search for tasklist in the issue body
 ISSUE_BODY_TASK_LIST_PATTERN = re.compile(
-    rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\s*```\[tasklist]\s*)"
-    rf"(?P<tasks>.*?)(?P<end>\s*```\s*{ISSUE_BODY_CLOSING_TAG})",
+    rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\s{{1,2}}```\[tasklist\])"
+    rf"(?P<tasks>.*?)"
+    rf"(?P<end>```\s{{1,2}}{ISSUE_BODY_CLOSING_TAG})",
     flags=re.DOTALL
 )
 
@@ -150,7 +151,7 @@ def build_issue_body(pr_number: int, body: str) -> str:
 
     # If the body already contains a tasklist, only add the new PR to the list
     return ISSUE_BODY_TASK_LIST_PATTERN.sub(
-        fr"\g<start>\g<tasks>\n- [ ] gh-{pr_number}\g<end>",
+        fr"\g<start>\g<tasks>- [ ] gh-{pr_number}\n\g<end>",
         body
     )
 
--- a/bedevere/util.py
+++ b/bedevere/util.py
@@ -1,7 +1,7 @@
 import enum
 import re
 import sys
-from typing import Any, Dict, Union
+from typing import Any, Dict
 
 import gidgethub
 from gidgethub.abc import GitHubAPI
@@ -25,7 +25,7 @@ PR_BODY_TEMPLATE = f"""\
 ISSUE_BODY_TAG_NAME = f"gh-linked-{PR}s"
 ISSUE_BODY_OPENING_TAG = f'<!-- {ISSUE_BODY_TAG_NAME} -->'
 ISSUE_BODY_CLOSING_TAG = f'<!-- /{ISSUE_BODY_TAG_NAME} -->'
-ISSUE_BODY_TEMPLATE = f"""\n
+ISSUE_BODY_TASK_LIST_TEMPLATE = f"""\n
 {ISSUE_BODY_OPENING_TAG}
 ```[tasklist]
 ### Linked PRs
@@ -145,7 +145,7 @@ def build_issue_body(pr_number: int, body: str) -> str:
 
     if not result:
         # If the body doesn't contain a tasklist, we add one using the template
-        body += ISSUE_BODY_TEMPLATE.format(pr_number=pr_number)
+        body += ISSUE_BODY_TASK_LIST_TEMPLATE.format(pr_number=pr_number)
         return body
 
     # If the body already contains a tasklist, only add the new PR to the list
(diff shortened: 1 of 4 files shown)


bedevere/util.py
```diff
@@ -1,23 +1,47 @@
...
+```[tasklist]
+### Linked PRs
+- [ ] gh-{{pr_number}}
+```
+{ISSUE_BODY_CLOSING_TAG}
 """
 
+# Regex pattern to search for tasklist in the issue body
+ISSUE_BODY_TASK_LIST_PATTERN = re.compile(
+    rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\s{{1,2}}```\[tasklist\])"
```
saadmk11, line 39: `\s{{1,2}}` is used instead of `\n` because when an issue is manually edited GitHub sometimes adds `\r` before `\n` so it becomes `\r\n`

bedevere/util.py
```diff
@@ -94,12 +118,50 @@ async def issue_for_PR(gh, pull_request):
...
+        tag_type=ISSUE,
+    )
+
+
+def build_issue_body(pr_number: int, body: str) -> str:
+    """Update the Issue body with related Pull Request."""
+    # If the body already contains a legacy closing tag
+    # (e.g. <!-- /gh-pr-number -->), then we use the legacy template
+    # TODO: Remove this when all the open issues using the legacy tag are closed
+    if PR_BODY_CLOSING_TAG.format(tag_type=PR) in body:
```
saadmk11, line 136: I am leaving the issues updated with the old implementation (e.g. `<!-- /gh-pr-number -->`) alone.

bedevere/util.py
ezio-melotti:
```suggestion
    rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\r?\n```\[tasklist\])"
```
Wouldn't this be better/more explicit then?
Alternatively you could also use `[\r\n]+` if you want to allow extra newlines.

bedevere/util.py
```diff
@@ -36,8 +36,9 @@
 
 # Regex pattern to search for tasklist in the issue body
 ISSUE_BODY_TASK_LIST_PATTERN = re.compile(
-    rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\s*```\[tasklist]\s*)"
-    rf"(?P<tasks>.*?)(?P<end>\s*```\s*{ISSUE_BODY_CLOSING_TAG})",
+    rf"(?P<start>{ISSUE_BODY_OPENING_TAG}\s{{1,2}}```\[tasklist\])"
+    rf"(?P<tasks>.*?)"
+    rf"(?P<end>```\s{{1,2}}{ISSUE_BODY_CLOSING_TAG})",
```
ezio-melotti, line 41:
```suggestion
    rf"(?P<end>```\r?\n{ISSUE_BODY_CLOSING_TAG})",
```

bedevere/util.py
saadmk11: I feel like this is better, wouldn't want the regex to be too greedy.

ezio-melotti:
This seems to work (at least with a single linked PR): https://github.com/python/cpython/issues/99443
We'll see if it works with multiple PRs too, but so far it looks good.

Thanks again @saadmk11!

saadmk11:
Hi @ezio-melotti, Looks like this is not working for multiple PRs (https://github.com/python/cpython/issues/96192). After some investigation I found that `tasklists` doesn't just render the new UI, it also re-writes the Issue body.

**Example:**

```markdown
<!-- gh-linked-prs -->

https://github.com/python/cpython/issues/96192#tasklist-block-8fd60492-131e-4b4a-abab-06800aa4bb11

<!-- /gh-linked-prs -->


<!-- gh-linked-prs -->

https://github.com/python/cpython/issues/96192#tasklist-block-ea9583fe-18a8-47ee-aec9-08a62427146b

<!-- /gh-linked-prs -->
```

I found this after I inspected the API response for https://github.com/python/cpython/issues/96192 (https://api.github.com/repos/python/cpython/issues/96192). This did not happen for the repositories I tested this PR on because it did not have this feature enabled.

This re-write causes multiple issues for us:

1. **Important** As you can see from the API response above the re-write even removes the reference to the linked PR (`gh-{pr-number}`). This causes the bot to add the same PR multiple times to the same issue (https://github.com/python/cpython/issues/96192) and we have no way to check what PR was added before.
2. The RegEx does not match and we can not be sure if the re-write behavior will remain after this feature out of beta. (All the other type of rendering features does not re-write issue body)

I would suggest switching to `checklists` at least for now until `tasklists` are out of beta. As I do not have access to any repository with this feature enabled I was not able to investigate any further.

ezio-melotti:
FWIW this is the relevant part of the body of the issue linked above, but as you noticed, this is different from what is received from the API.
``````

<!-- gh-linked-prs -->

```[tasklist]
### Linked PRs
- [x] https://github.com/python/cpython/pull/96194
```


<!-- /gh-linked-prs -->


<!-- gh-linked-prs -->

```[tasklist]
### Linked PRs
- [x] https://github.com/python/cpython/pull/96194
```


<!-- /gh-linked-prs -->


<!-- gh-linked-prs -->

```[tasklist]
### Linked PRs
- [ ] https://github.com/python/cpython/pull/99455
```


<!-- /gh-linked-prs -->

``````

> we have no way to check what PR was added before.

The `gh-XXXXX` expansion seems straightforward -- can't you just look for `https://github.com/python/cpython/(?:pull|issue)/\d+`  (or even `(?:https://github.com/python/cpython/(?:pull|issue)/|gh-)\d+`) instead?

> I would suggest switching to checklists at least for now until tasklists are out of beta. 

This is definitely an option if the above suggestion fails.

> As I do not have access to any repository with this feature enabled I was not able to investigate any further.

You can join the waitlist to request tasklists being enabled on one of your repos here: https://github.com/features/issues/signup

I also contacted GitHub for feedback, so if the suggestion doesn't work we can switch to checklist, and reconsider tasklists if/when they offer a solution.

saadmk11:
> > we have no way to check what PR was added before.
> 
> The `gh-XXXXX` expansion seems straightforward -- can't you just look for `https://github.com/python/cpython/(?:pull|issue)/\d+` (or even `(?:https://github.com/python/cpython/(?:pull|issue)/|gh-)\d+`) instead?

Not sure what you mean here. The API returns this: `https://github.com/python/cpython/issues/96192#tasklist-block-8fd60492-131e-4b4a-abab-06800aa4bb11`. Here the URL is not the URL of the pull request but the URL of the issue the API is requesting.

> > As I do not have access to any repository with this feature enabled I was not able to investigate any further.
> 
> You can join the waitlist to request tasklists being enabled on one of your repos here: https://github.com/features/issues/signup

Unfortunately you can't join the wait-list as an individual user, only GitHub organizations are accepted.

ezio-melotti:
> Here the URL is not the URL of the pull request but the URL of the issue the API is requesting.

I see, I thought it was the full URL for the PRs.

In that case, let's just switch to checklists for now.

ezio-melotti:
> Seems like auto marking a checklist items as completed (e.g: - [x] gh-170) when a PR is merged does not work

If that's the case (i.e. merging a PR doesn't mark it as completed in the checklist), then there is no advantage in using checklists over plain lists.  The difference between checklists and plain lists is that the checklist shows the number of "tasks" at the top of the issue (as `(1 of 3 tasks)`) and that the PR will show up as `Tracked in #XXX`, however the former is useless if merged PRs are not marked as completed, and the latter is already covered by bedevere adding the issue link to each PR.

IOW, it's probably better to go back to plain lists for the time being.

omerbensaadon:
Hey folks, we're investigating this as a bug on our end. So sorry for the delay here, we'll try and get this fixed in the new year and making sure multiple PRs work in these blocks. 

Again, so sorry you all had to wait this long for a response and thank you so much for being Alpha (now Beta!) users :)

CAM-Gerlach:
Thanks @omerbensaadon ! Looking forward to it!

PR #547, opened 2023-04-30
Remove codecov dependency; skip running CI on 3.12 for now

AlexWaygood:
The `codecov` package is no longer installable, meaning that `pip install dev-requirements.txt` for this repo now always fails. It also doesn't seem to be necessary to install it in CI; we don't use it for anything as far as I can see, and the codecov action doesn't require it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,10 +13,11 @@ jobs:
 
     strategy:
       matrix:
-        # TODO replace 3.12.0-alpha.4 with 3.12-dev when
+        # TODO add 3.12.0-dev when
         # aiohttp, frozenlist and yarl support alpha 5+
         # https://github.com/python/blurb_it/pull/330#issuecomment-1449496275
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12.0-alpha.4"]
+        # https://github.com/python/bedevere/pull/547#issuecomment-1529153851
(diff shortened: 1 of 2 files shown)


AlexWaygood:
It looks like the tests are failing on 3.12, but in actual fact, we never get as far as running the tests! `tox` tries to use `virtualenv` to set up a virtual environment, but `virtualenv` crashes. Why? It tries to pass in an `onexc` argument to `shutil.rmtree`, because `virtualenv` sees that we're running on Python 3.12, and wants to avoid a `DeprecationWarning` (since the old `onerror` argument to `rmtree` is deprecated in Python 3.12). _But_ we pin 3.12.0a4 following https://github.com/python/bedevere/commit/5f61b8877f8d18f1d7a0b32211ce3eb964acd67a, meaning the `onexc` argument to `shutil.rmtree` doesn't exist yet (it was added in 3.12.0a7, in https://github.com/python/cpython/commit/7f760c2fca3dc5f46a518f5b7622783039bc8b7b).

Not sure what the best solution here is! The original reasons for pinning the 3.12 workflow to 3.12.0a4 in #542 still apply -- `aiohttp` and `frozenlist` still don't support being installed on Python 3.12.0a5 or higher.

Maybe we should just skip the 3.12 workflow altogether for now?

hauntsaninja:
Yeah, I'd skip the 3.12 workflow for now

PR #553, opened 2023-05-01
CI: set `fail-fast` to False

AlexWaygood:
With `fail-fast` set to True (the default), all running jobs in the matrix are cancelled as soon as a single one fails. This makes it hard to see exactly which Python versions a PR is failing on, if there are failures, because the CI just becomes a sea of red, even if the PR is only failing on a single Python version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,13 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
         # TODO add 3.12-dev when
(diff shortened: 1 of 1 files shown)


hugovk (approved):
👍 

This helps debug failures.

A little cosmetic suggestion: I tend to put `fail-fast: false` between `strategy:` and `matrix:`, maybe following the lead of the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast), and consistency makes it easier to diff files between projects.

Issue #554, opened 2023-05-01
Bedevere shouldn't add "awaiting review" label on draft PRs


itamaro:
I noticed that when opening draft PRs on the cpython repo, Bedevere still goes ahead and adds the "awaiting review" label.
I don't want to spam with review requests for PRs that are not ready for review (hence "draft"), so I'd like to suggest modifying the state machine to acknowledge the draft state (e.g. don't add any labels until the PR is published).

itamaro:
I'm happy to try to make this change myself (maybe with some guidance as I never touched this codebase before :) )

Mariatta:
I think the main change is on here:

https://github.com/python/bedevere/blob/main/bedevere/stage.py#L104
Here, we're handling the "pull_request.opened" webhook event, and apply either the `awaiting review` or `awaiting core review` label.

Since the webhook event sends the `draft` status as True or False, we can use that to determine whether or not we should apply the label when the PR is opened.

```
if event.data["draft"] is False:
# apply label
```

We should also add apply the label when the PR was changed from `draft` to `ready to review`. So we may need to add the decorator `@router.register("pull_request", action="edited")`.

Mariatta:
And I guess we should also handle the situation when the PR was changed from "ready to review" back to the "draft" state.

PR #555, opened 2023-05-01
gh-554: Handle Draft PRs

itamaro:
1. if opened PR is draft - don't add labels
2. if draft PR is published - add labels as a new non-draft PR
3. if a PR is unpublished - clear labels

### DIFF
--- a/tests/test_stage.py
+++ b/tests/test_stage.py
@@ -72,6 +72,46 @@ async def test_stage():
     assert post_[1] == [awaiting.Blocker.merge.value]
 
 
+async def test_opened_draft_pr():
+    # New Draft PR from a core dev.
+    username = "brettcannon"
+    issue_url = "https://api.github.com/issue/42"
+    data = {
+        "action": "opened",
+        "pull_request": {
+            "user": {
+                "login": username,
+            },
+            "issue_url": issue_url,
+            "draft": True,
+        },
+    }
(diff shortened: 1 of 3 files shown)


Mariatta:
Thank you!
Can you add test for when a PR went from published to draft?

itamaro:
> Can you add test for when a PR went from published to draft?

Sure!

What's the correct way to mock an existing PR with labels and assert they get deleted?
It looks like the labels are associated with "issue data" based on other tests

Mariatta:
We would test that there is an API call for removing the labels from PR.

itamaro:
updated the PR - is this what you meant? (asserting on `gh.delete_url`)

tests/test_stage.py
```diff
@@ -72,6 +72,74 @@ async def test_stage():
...
+    )
+    await awaiting.router.dispatch(event, gh)
+    assert len(gh.post_) == 0
+    assert (
+        gh.delete_url ==
+        f"https://api.github.com/repos/python/cpython/issues/12345/labels/{encoded_label}"
+    )
+
+
 async def test_opened_pr():
```
Mariatta, line 143: Can you adjust this test case, the `data` dict should perhaps have `"draft": False` value

itamaro: done!

Mariatta (approved):
Thank you!

Issue #556, opened 2023-05-06
Renaming issues/PRs incorrectly changes `awaiting change review` and `awaiting merge` to `awaiting review`


arhadthedev:
While changing `bpo-` to `gh-` I noticed that Bedevere makes forbidden transitions in response:

- <kbd>awaiting change review</kbd> → <kbd>awaiting review</kbd>: python/cpython#28728
- <kbd>awaiting merge</kbd> → <kbd>awaiting review</kbd>: python/cpython#20310

According to README, `awaiting review` has no incoming conditions:

```mermaid
flowchart TD
    A([Published PR]):::creator
    A_draft([New Draft PR]):::creator
    A_draft -- publish draft by contributor --> A:::creator
    A -- by contributor --> B[Awaiting review]:::anyone
    A -- by core dev --> C[Awaiting core review]:::coredev
    B & C -- new review by\nanother contributor --> C
    C & B & E  -- new core review\nrequests changes --> D[Awaiting changes]:::creator
    D -- changes by contributor --> E[Awaiting change review]:::coredev
    C & E & B -- new core review\napproves ---> F[Awaiting merge]:::coredev
    G[When a review is dismissed\nthe highest remaining state is restored\nbased on the remaining reviews]
classDef creator stroke:#CC0;
classDef anyone stroke:#00C;
classDef coredev stroke:#0C0;
classDef triager stroke:#C0C;
linkStyle 0,1,7 stroke:#CC0,color:auto;
linkStyle 2,3 stroke:#00C,color:auto;
linkStyle 4,5,6,8,9,10 stroke:#0C0,color:auto;
```

I suspect the bug is here, in a handler of PR/issue modifications introduced in gh-555: https://github.com/python/bedevere/blob/0f350cee365b615a1c4fefd0e2cf77e6215e4164/bedevere/stage.py#L127-L131

We need to check whether assignment of `Blocker.core_review` or `Blocker.review` is allowed before executing the `else` branch.

*Ideally (for some distant future) we need to move tag management into a separate `.py` file that reads existing tags, checks if a new one is applicable, and removes the old ones. However, I have no free time to do it currently (because it requires massive modification of both reactors and their tests).*

cc @itamaro

itamaro:
oops, you're right, this handler is doing way more than it was intended!
that handler should respond only to PR edits that change the PR from draft to published or from published to draft.

arhadthedev:
@itamaro Could you submit a PR, please?

itamaro:
[...]

PR #557, opened 2023-05-06
gh-556: Fix over-reaching PR-edited event handler

itamaro:
gh-555 introduced the PR edited handler to handle draft PRs. The implementation was over-reaching, and triggered on any edited events (including renaming PR). This PR fixes that by removing the PR-edited handler and replacing it with two more specific handlers that more narrowly match the original intent, namely, respond to changes in the draft state of the PR (publish / unpublish).

Fixes gh-556

### DIFF
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -101,15 +101,13 @@ async def stage(gh, issue, blocked_on):
 
 
 async def stage_for_review(gh, pull_request):
-    """Apply awaiting review labels."""
+    """Apply "awaiting review" label."""
     issue = await util.issue_for_PR(gh, pull_request)
     username = util.user_login(pull_request)
-    blocked_on = (
-        Blocker.core_review
-        if await util.is_core_dev(gh, username)
-        else Blocker.review
-    )
-    await stage(gh, issue, blocked_on)
+    if await util.is_core_dev(gh, username):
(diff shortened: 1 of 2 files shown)


bedevere/stage.py
```diff
@@ -100,6 +100,18 @@ async def stage(gh, issue, blocked_on):
     await gh.post(issue["labels_url"], data=[label_name])
 
 
+async def stage_for_review(gh, pull_request):
+    """Apply awaiting review labels."""
```
ezio-melotti, line 104:
```suggestion
    """Apply "awaiting review" labels."""
```

bedevere/stage.py
```diff
@@ -100,6 +100,18 @@ async def stage(gh, issue, blocked_on):
...
+    issue = await util.issue_for_PR(gh, pull_request)
+    username = util.user_login(pull_request)
+    blocked_on = (
+        Blocker.core_review
+        if await util.is_core_dev(gh, username)
+        else Blocker.review
+    )
```
ezio-melotti, line 111:
```suggestion
    if await util.is_core_dev(gh, username):
        blocked_on = Blocker.core_review
    else:
        blocked_on = Blocker.review
```
I think a regular `if/else` would read better here.

JelleZijlstra (approved):

PR #583, opened 2023-09-08
Only comment about new commits if the PR is still open.

Mariatta:
This should fix the issue where the "awaiting core review" label gets added after PR has ben merged. Example: https://github.com/python/cpython/pull/109082

### DIFF
--- a/bedevere/stage.py
+++ b/bedevere/stage.py
@@ -147,14 +147,16 @@ async def new_commit_pushed(event, gh, *arg, **kwargs):
         # get the latest commit hash
         commit_hash = commits[-1]["id"]
         pr = await util.get_pr_for_commit(gh, commit_hash)
-        for label in util.labels(pr):
-            if label == "awaiting merge":
-                issue = await util.issue_for_PR(gh, pr)
-                greeting = "There's a new commit after the PR has been approved."
-                await request_core_review(
-                    gh, issue, blocker=Blocker.core_review, greeting=greeting
-                )
-                break
+        if pr["state"] != "closed":
+            # only process the event if PR is still open
+            for label in util.labels(pr):
+                if label == "awaiting merge":
+                    issue = await util.issue_for_PR(gh, pr)
+                    greeting = "There's a new commit after the PR has been approved."
+                    await request_core_review(
+                        gh, issue, blocker=Blocker.core_review, greeting=greeting
+                    )
+                    break
 
 
 async def core_dev_reviewers(gh, pull_request_url):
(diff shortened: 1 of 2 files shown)


AA-Turner:
I wonder if the entire ``new_commit_pushed`` should be removed? It doesn't seem to work even in the cases that it should -- Paul approved [GH-109130](https://github.com/python/cpython/pull/109130), putting it into `awaiting merge`, and I've pushed several commits since, which haven't triggered Bedevere to change the state.

A

AlexWaygood:
I agree with Adam. Even if this logic could be fully fixed, I'm not sure it's worth fixing it. I can't think of a situation in which I'd find this comment from Bedevere useful, since I already get a notification on GitHub if a new commit has been pushed on a PR I've already reviewed. I feel like the extra comment from Bedevere would just be annoying and lead to me having to delete two emails from my inbox, rather than just one.

Mariatta:
Hmm that's interesting, looking at the PR mentioned by @AA-Turner , it looks like the new commits triggered the `synchronize` event instead.

I'll investigate this some more.

I'm ok with removing bedevere's comment, but I think we still need to logic to reapply the `[awaiting core review]` label.

AA-Turner:
> I think we still need to logic to reapply the `[awaiting core review]` label

I'd push back slightly in line with Alex's thoughts and ask *why* here: Bedevere seems to have only started applying the label [three weeks ago](https://github.com/python/cpython/issues?q=%22new+commit+after+the+PR+has+been+approved%22+sort%3Aupdated-asc), so from the time CPython moved to GitHub to ~mid August 2023 the logic hasn't been running and (to my knowledge) no one has complained about it.

A

AlexWaygood:
> I think we still need to logic to reapply the `[awaiting core review]` label.

I don't have a strong opinion on this question, but it feels strange to me that merging in `main` (for example) would cause Bedevere to remove the "awaiting merge" label. Personally, I think it would be fine for the "awaiting merge" to stay there unless a core dev requests changes, adds "DO-NOT-MERGE", or dismisses their previous approving review

pradyunsg:
> it looks like the new commits triggered the `synchronize` event instead.

Indeed.

Currently, we have `@router.register("push")` which is triggered for ([docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push)):

> when there is a push to a repository branch

This means it'll trigger for pushes to `main` on python/cpython (or any other branch in python/cpython) but it won't trigger when PRs are filed from a branch that doesn't live in python/cpython.

We should change this to use the `pull_request` event ([docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)):

```py
@router.register("pull_request", action="opened")
@router.register("pull_request", action="reopened")
@router.register("pull_request", action="synchronize")
```

IIUC, those are all the events that we'd want to listen to for determining whether the label needs to be updated.

PR #593, opened 2023-10-11
Tweak the language for the help text

### LINKED ISSUE #590: Update NEWS message
The current NEWS message reads like this:

```
Most changes to Python [require a NEWS entry](https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python).

Please add it using the [blurb_it](https://blurb-it.herokuapp.com/) web app or the [blurb](https://pypi.org/project/blurb/) command-line tool.
```

I see many contributors then add a NEWS entry on PRs where they don't need it. This is understandable, but it would be nice if the message could suggest something like "If this change has little impact on Python users, you can wait for a maintainer to confirm that a NEWS entry is required. Maintainers are able to bypass this check."

That would likely help out contributors [like this one](https://github.com/python/cpython/pull/110566) where the NEWS entry is unnecessary but the PR checks are failing _because_ the contributor added an entry. Rather than have them stress out about learning enough ReST syntax, it would be nice if they were given permission to wait for us to tell them whether it's necessary or not.

pradyunsg:
The existing message did not cover the option of skipping the entry with `skip news`, which can result in unnecessary work being done by new contributors.

Resolves #590

### DIFF
--- a/bedevere/news.py
+++ b/bedevere/news.py
@@ -29,10 +29,13 @@ SKIP_LABEL_STATUS = create_status(util.StatusState.SUCCESS,
 
 HELP = f"""\
 Most changes to Python [require a NEWS entry]\
-(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python).
+(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python). \
+You can add a NEWS entry using the [blurb_it]({BLURB_IT_URL}) web app or \
+the [blurb]({BLURB_PYPI_URL}) command-line tool.
 
-Please add it using the [blurb_it]({BLURB_IT_URL}) web app or the [blurb]\
-({BLURB_PYPI_URL}) command-line tool."""
+If this change has little impact on Python users, wait for a maintainer to confirm \
+that a NEWS entry is not required (they can apply the `skip news` label).
+"""
 
 
 async def check_news(gh, pull_request, files=None):
--- a/bedevere/news.py
+++ b/bedevere/news.py
@@ -30,11 +30,11 @@ SKIP_LABEL_STATUS = create_status(util.StatusState.SUCCESS,
 HELP = f"""\
 Most changes to Python [require a NEWS entry]\
 (https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python). \
-You can add a NEWS entry using the [blurb_it]({BLURB_IT_URL}) web app or \
+Add one using the [blurb_it]({BLURB_IT_URL}) web app or \
 the [blurb]({BLURB_PYPI_URL}) command-line tool.
 
-If this change has little impact on Python users, wait for a maintainer to confirm \
-that a NEWS entry is not required (they can apply the `skip news` label).
+If this change has little impact on Python users, wait for a maintainer to \
+apply the `skip news` label instead.
 """
 
 


ezio-melotti (approved):

pradyunsg:
> currently we're linking to those docs, but we're not saying why we're linking to those docs!

Indeed.

I think the devguide link doesn't cover things in sufficient detail and links to content that isn't really written for contributors (the page is "for core developers").

I do consider this a separate issue though from the one we're resolving with this -- this is mentioning information that isn't mentioned in the relevant page on the devguide so there isn't really a duplication concern and we should improve the devguide separately.

> Another, separate, thing we should do is figure out if there's a way to get Bedevere to only post this comment once on any given PR.

I agree, though as you say, this should be separate. :)

AlexWaygood:
> I think the devguide link is to content that isn't really written for contributors (the page is "for core developers").
> 
> I do consider this a separate issue though from the one we're resolving with this -- this is mentioning information that isn't mentioned in the relevant page on the devguide so there isn't really a duplication concern and we should improve the devguide separately.

Okay, fair enough! I do think concision is really important with this kind of thing, though, or people just switch off and don't read the thing. I'll try to suggest something in the morning :)

bedevere/news.py
```diff
@@ -29,10 +29,13 @@
 
 HELP = f"""\
 Most changes to Python [require a NEWS entry]\
-(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python).
+(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python). \
+You can add a NEWS entry using the [blurb_it]({BLURB_IT_URL}) web app or \
```
AlexWaygood, line 33:
```suggestion
Add one using the [blurb_it]({BLURB_IT_URL}) web app or \
```

bedevere/news.py
```diff
@@ -29,10 +29,13 @@
...
+the [blurb]({BLURB_PYPI_URL}) command-line tool.
 
-Please add it using the [blurb_it]({BLURB_IT_URL}) web app or the [blurb]\
-({BLURB_PYPI_URL}) command-line tool."""
+If this change has little impact on Python users, wait for a maintainer to confirm \
+that a NEWS entry is not required (they can apply the `skip news` label).
```
AlexWaygood, line 37:
```suggestion
If this change has little impact on Python users, wait for a maintainer to \
apply the `skip news` label instead.
```

AlexWaygood (commented):
Some suggestions to make the language slightly more concise:

AlexWaygood (approved):
Thanks!

ezio-melotti (approved):

bedevere/news.py
```diff
@@ -29,10 +29,13 @@
...
 Most changes to Python [require a NEWS entry]\
-(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python).
+(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python). \
+Add one using the [blurb_it]({BLURB_IT_URL}) web app or \
+the [blurb]({BLURB_PYPI_URL}) command-line tool.
 
-Please add it using the [blurb_it]({BLURB_IT_URL}) web app or the [blurb]\
-({BLURB_PYPI_URL}) command-line tool."""
+If this change has little impact on Python users, wait for a maintainer to \
+apply the `skip news` label instead.
```
hugovk, line 37: Flip the order, as trivial changes are more likely for new contributors: 

```suggestion
If this change has little impact on Python users, wait for a maintainer to \
apply the `skip news` label.

Otherwise, [add a `NEWS` entry]\
(https://devguide.python.org/committing/#updating-news-and-what-s-new-in-python) \
using the [blurb_it]({BLURB_IT_URL}) web app or \
the [blurb]({BLURB_PYPI_URL}) command-line tool.
```

pradyunsg: Discussed this in person and we decided to keep things as-is for now.

PR #594, opened 2023-10-12
Add logging to debug bedevere label staging

Mariatta:


pradyunsg (commented):
LGTM, although we should make sure we revert this since the debug logs might be contain information that would be suboptimal to be logging.

PR #596, opened 2023-10-13
Add linting and auto-formatting via pre-commit

pradyunsg:
This would make the codebase formatting consistent and help avoid discussions about the formatting on PRs.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,1 @@
+35235fd609a52db835beb2a96d0da461a0eab3ed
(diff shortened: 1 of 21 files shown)


pradyunsg:
For the last commit, it will only work if we don't do a squash merge. If we're going to do a squash merge, we should remove it.

pradyunsg:
And, I've made a request to request installation of pre-commit.ci, to get autofixes on PRs, which is something that @Mariatta expressed interest in.

Mariatta (approved):

Mariatta:
Thank you. I think the changes look good. I don't have the necessary permission to do the merge commit though.

PR #617, opened 2024-01-19
Add the "docs" label also for Doc/*/.nitignore

### LINKED ISSUE #616: Add docs label for PRs that touch Doc/tools/.nitignore
See e.g. https://github.com/python/cpython/pull/114280 or https://github.com/python/cpython/pull/114194

skirpichev:
Closes #616

### DIFF
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -43,7 +43,9 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         if util.is_news_dir(filename):
             news = True
         filepath = pathlib.PurePath(filename)
-        if filepath.suffix == ".rst":
+        if filepath.suffix == ".rst" or (
+            util.is_docs_dir(filename) and filepath.name == ".nitignore"
+        ):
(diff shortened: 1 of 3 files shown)


bedevere/prtype.py
```diff
@@ -43,7 +43,9 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         if util.is_news_dir(filename):
             news = True
         filepath = pathlib.PurePath(filename)
-        if filepath.suffix == ".rst":
+        if filepath.suffix == ".rst" or (
+            util.is_docs_dir(filename) and filepath.name == ".nitignore"
+        ):
```
hugovk, line 48: We only have one file named this, and I don't expect we'll add others (if we do, we can deal with it then):

```suggestion
        if filepath.suffix == ".rst" or filepath.name == ".nitignore":
```

hugovk (approved):
Thanks!

AA-Turner (approved):

PR #632, opened 2024-03-01
Group dependabot updates

hugovk:
Open one grouped PR per month instead of many, to help deal with notification fatigue. 

Compare first two repos with the last:

image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,10 @@ updates:
       interval: monthly
     assignees:
       - "ezio-melotti"
-    open-pull-requests-limit: 10
+    groups:
+      pip:
+        patterns:
+          - "*"
 
   # Maintain dependencies for GitHub Actions
   - package-ecosystem: "github-actions"
@@ -16,4 +19,7 @@ updates:
       interval: monthly
     assignees:
       - "ezio-melotti"
-    open-pull-requests-limit: 10
+    groups:
+      actions:
+        patterns:
+          - "*"


itamaro (approved):
neat!

hauntsaninja (approved):
Oh thank god

ezio-melotti:
FWIW I don't mind dealing with them separately, which I find generally easier than a single PR with multiple updates.  I'm also auto assigned to them and I generally review and merge them all within a day (as you can see from the screenshot).  I understand that this still creates a number of notifications for other people, but they can generally ignore them and/or mark them as "done", especially if I already reviewed and merged them.

Unfortunately [@dependabot will always submit *monthly* updates on the first day of the month](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval), and that results in a large number of notification on the first of each month.  For weekly updates it's possible to set the day of the week -- if it was possible to set the day for monthly releases too we might have been able to stagger the PR from different repos on different days.

hugovk:
In just the last day, I've received 21 Dependabot notifications from 7 repos (across 4 organisations).

Also in one day, from Renovate (similar to Dependabot) I've received 17 notifications from 8 repos (across 5 orgs).

(btw Renovate allows selecting day-of-week staggering monthly updates. We've set Pillow repos to trigger on different days, and avoid tying up the CI on potential release days.)

Luckily today was the first day of the month and not the first of the quarter, or I would have had a raft of pre-commit updates too.

Instead of 38 notifications, if we can have these closer to 15, it would be much better. Each of these notifications pushes _others_ further down the list. Even if it's easy to dismiss, they all add to "notification/alert fatigue" that can contribute to burnout. (I've seen some projects turn off Dependabot because of the perceived noise, but I'm not suggesting that.)

Having separate PRs uses up more CI time: often when merging one, the next in the list is rebased, meaning you need to wait another couple of minutes. Repeat x 5. Having so many PRs in the tracker can also make it harder to search for things later.

We have a bit of a special case about Codecov (let's talk about how to handle that another time), but usually it doesn't really matter if they're one-by-one or all-at-once. Sometimes they need to be done together (the recent actions/upload-artifact and download-artifact updates).

See also https://github.com/python/python-docs-theme/pull/170#discussion_r1477356230.

hugovk:
PS Between starting writing and posting that, an extra 5 Dependabot notifications have come in from 2 more repos.

PR #657, opened 2024-12-06
Add docs label for markdown files

### LINKED ISSUE #656: Treat InternalDocs/* like Docs/*?
I.e. add label "docs" (or similar), "skip news", etc.

An example: https://github.com/python/cpython/pull/127687

skirpichev:
Closes #656

### DIFF
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -43,7 +43,7 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         if util.is_news_dir(filename):
             news = True
         filepath = pathlib.PurePath(filename)
-        if filepath.suffix in [".md", ".rst"] or filepath.name == ".nitignore":
+        if filepath.suffix in {".md", ".rst"} or filepath.name == ".nitignore":
             docs = True
         elif filepath.name.startswith(("test_", "_test")):
             tests = True
--- a/bedevere/prtype.py
+++ b/bedevere/prtype.py
@@ -43,7 +43,7 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         if util.is_news_dir(filename):
             news = True
         filepath = pathlib.PurePath(filename)
-        if filepath.suffix in [".rst", ".md"] or filepath.name == ".nitignore":
+        if filepath.suffix in [".md", ".rst"] or filepath.name == ".nitignore":
             docs = True
         elif filepath.name.startswith(("test_", "_test")):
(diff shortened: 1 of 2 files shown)


bedevere/prtype.py
```diff
@@ -43,7 +43,7 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         if util.is_news_dir(filename):
             news = True
         filepath = pathlib.PurePath(filename)
-        if filepath.suffix == ".rst" or filepath.name == ".nitignore":
+        if filepath.suffix in [".rst", ".md"] or filepath.name == ".nitignore":
```
hugovk, line 46: Let's put them in alphabetical order:
```suggestion
        if filepath.suffix in [".md", ".rst"] or filepath.name == ".nitignore":
```

hugovk (approved):
Thanks!

bedevere/prtype.py
```diff
@@ -43,7 +43,7 @@ async def classify_by_filepaths(gh, pull_request, filenames):
         if util.is_news_dir(filename):
             news = True
         filepath = pathlib.PurePath(filename)
-        if filepath.suffix == ".rst" or filepath.name == ".nitignore":
+        if filepath.suffix in [".md", ".rst"] or filepath.name == ".nitignore":
```
AA-Turner, line 46: Tiny optimisation (the optimiser might catch this, but stylisticly I'd prefer avoiding membership tests with lists)
```suggestion
        if filepath.suffix in {".md", ".rst"} or filepath.name == ".nitignore":
```

AA-Turner (approved):

tests/test_prtype.py
```diff
@@ -66,7 +66,7 @@ async def test_news_only():
 
 
 async def test_docs_no_news():
-    filenames = {"path/to/docs1.rst"}
+    filenames = {"path/to/docs1.rst", "other/path/to/docs2.md"}
```
ezio-melotti, line 69: Wouldn't this test pass even without the changes in `prtype.py`?
The filenames include an `.rst` file, so the `docs` label should be applied regardless.
I would suggest to verify this, and possibly turn the list of filenames into a fixture.

skirpichev: > Wouldn't this test pass even without the changes in prtype.py?

No.  In this case classify_by_filepaths() will quickly return an empty list, walking over filenames (else branch).

